PR #8, opened 2019-11-07
linguistic edits in Garima, moved records from new to numbered folder

thea-m:


thea-m:
and created new records

PietroLiuzzo:
as we are still in a testing phase, may I suggest that moving files is not a significative change which needs a PR? such changes could be simply pushed to the master branch by those who can do that.

DenisNosnitsin1970:
ok.

thea-m:
sure, but I had the impression that we are already having quite some self-merging - I think that (for me), it is easier to make PRs as a rule instead of wondering whether I should do so or not, and I would not mind (quickly, obviously) reviewing such PRs for others.

PR #13, opened 2019-11-20
Update PRS3587Dionysiu.xml

eu-genia:


thea-m:
Let's hope no confusion with PRS7973PseudoD will arrise
(I am a fan of quickly adding wikidata whenever working on persons who have it, but it's of course not necessary.)

PR #14, opened 2019-11-20
substituted "place" by "person" in change elemetn

thea-m:
Or do we have a policy about changing other people's change elements?


eu-genia:
this was the result of the batch conversion: absolutely all automatically created records have this line. maybe we either leave it or Pietro at some point can automatically correct? @PietroLiuzzo

thea-m:
I think that only the person records had the wrong entity mentioned.

PietroLiuzzo:
yes, this sort of changes should be listed for a task during data freeze before a release, because here you are changing 9000+ files. You do not really expect me to review them all, isn't it?

thea-m:
No, do what seems best to you.

PietroLiuzzo:
just to specify why this is best done during  a freeze, sending this  9000+ change to the server will CERTAINLY kill it and result in inconsistencies

thea-m:
So will you take care of it in the next freeze?

PietroLiuzzo:
yes

thea-m:
@sologebre I think you didn't intend to merge your branch into this one. Could you commit them to a new branch?

PR #45, opened 2020-01-23
reproduced changes previously in PR 41

PietroLiuzzo:


PietroLiuzzo:
@DenisNosnitsin1970 these are the changes which were in the Petersburg branch and I have removed to be able to merge it. so, they are your changes, it should be easy to review them ;)

PR #69, opened 2020-03-03
persons for BL orient679

thea-m:


thea-m:
The record is that of ʾIyāsu III (https://betamasaheft.eu/persons/PRS5618IyasuII/main), the ID was created automatically (and a bit unfortunately in this case)

DenisNosnitsin1970:
Can we correct it? It is very misleading.

thea-m:
hmm, I'm not sure. We could of course, but we would then have to check all occurrences etc. etc. I'm quite sure I saw other records with the same phenomenon (maybe also patriarchs or metropolitans)... I think I would rather not do it currently, but you can of course :)

PR #86, opened 2020-03-16
Transliteration2

DariaElagina:


thea-m:
The conflict in the header seems to me a question of layout - are there any changes there @DariaElagina ? If not, the original header could be kept in all files

PR #92, opened 2020-03-24
Persons for bl corrections

eu-genia:


PRS4001-5000/PRS4822Gregory.xml
```diff
@@ -45,18 +45,23 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
       <body>
          <listPerson>
-            <person sex="1">
-               <persName>
-                  
-                  Gregory of Nazianzos
-               </persName>
-               <persName type="alt">Gorgoryos</persName>
-               <!-- St -->
+            <person sex="1" sameAs="">
```
thea-m: wikidata :)

eu-genia: oops… :) done

eu-genia: just pushed another item sorry confused branches (too many… and my home connection is a bit dreamy)

new/PRS12917JohnRome.xml
```diff
@@ -0,0 +1,58 @@
...
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="en" xml:id="n1">John of Rome</persName>
+                    <persName xml:lang="en" xml:id="n2">John of the Golden Gospel</persName>
+                    <faith type="Coptic"/>
+
+                    <death>He died on 16 Abib and is commemorated in the Coptic Church on this date</death>
```
thea-m: it could be marked up with `<date ref="ethiocal:Hamle16">`, since it also true for the Ethiopian tradition, then the comment with the link could be deleted, I think.

PRS4001-5000/PRS4822Gregory.xml
thea-m: Ah, I know the problem :)

new/PRS12917JohnRome.xml
eu-genia: ok done thanks

PR #97, opened 2020-03-27
Scribes e s

eu-genia:


DenisNosnitsin1970:
Conflict?

PietroLiuzzo:
I took the branch version to be better than the master. hope it was ok, there is no more conflict. you are also welcome to use the very nice conflict resolution editor offered here.

PR #120, opened 2020-04-14
Gabra iwm

DenisNosnitsin1970:
New record. A later owner of half-same name as the scribe.


thea-m:
Does 
```
A monk from <placeName ref="INS0479DabraM">Dabra Māryām
                      Qoḥayn of ʾabbā ʾAbsādi</placeName>
```
refer to this person or another?

PR #122, opened 2020-04-15
New entry walda maryam

MassimoVilla:
An individual mentioned in one Casamari MS (possibly owner).


new/PRS12962WaldaMa.xml
```diff
@@ -0,0 +1,59 @@
...
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ወልደ፡ ማርያም፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Walda Māryām</persName>
+                    <faith type="EOTC"/>     
+                    <floruit from="1882" to="1889"/>             
+                </person>
+                <listRelation>
+                    <relation name="saws:hasOwned" active="PRS12962WaldaMa" passive="MNC009" evidence="conjecture"/>
```
thea-m, line 54: The @evidence attribute is currently not employed so freely (as far as I remember, it is only used for dates). If there is doubt about his ownership, it might be clearer to either have a `<desc>` in the `<relation>` and describe the situation there, and/or to use the `<certainty>` element (having a sentence in `<desc>`  always makes it easier for users).  
https://betamasaheft.eu/Guidelines/?id=certainty

PR #123, opened 2020-04-16
schwa

thea-m:


eu-genia:
@DariaElagina I noted most of these were created by you, there must be something to check in your diacritics shortcuts.

PR #145, opened 2020-05-07
added name

thea-m:
It seems like it is him:
https://books.google.de/books?id=DgJekXyACmgC&pg=PP11&lpg=PP11&dq=petros+gebre+selassie+makarios&source=bl&ots=hmx-gJt0uL&sig=ACfU3U1fAmB8iLm9ruR38fsUuuZHMkCw3Q&hl=de&sa=X&ved=2ahUKEwjcoLjq1KHpAhWQjKQKHXXPCl0Q6AEwAHoECAoQAQ#v=onepage&q=petros%20gebre%20selassie%20makarios&f=false
https://www.csmonitor.com/1992/0115/15112.html

I am not sure what source to give in the record. What do you think?

### DIFF
--- a/PRS11001-12000/PRS11985PetrosGS.xml
+++ b/PRS11001-12000/PRS11985PetrosGS.xml
@@ -34,7 +34,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
         <profileDesc>
             <creation/>
             <abstract>
-                <p/>
+                <p>Bishop Makarios is bishop of the United States of the Eritrean Orthodox Church and a member of the Coptic Orthodox Synod.</p>
             </abstract>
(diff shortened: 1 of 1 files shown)


DenisNosnitsin1970:
It is a difficult case, as far as I can see, he is aged but I think he is still alive and is looking at you stricktly but with great care, 
https://eritrea-focus.org/abune-makarios-president-isaias-must-consult-the-eritrean-people/.
Moreover, he is a public person. The reference should be kept, but it is of course not the true reference. I think we should not necessarily identify him  in the printed article as Makarios (it is not in the ms, isn't it?).

thea-m:
It is in the Arabic note which states that the text was translated by ʾanbā Makāriyūs (but not equating him with Petros)

DenisNosnitsin1970:
Then we are not obliged to equate them either.

PR #175, opened 2020-06-24
new entries for scribes

DariaElagina:


DariaElagina:
@thea-m  I think you are right! I have not thought of this possinility, thank you! I will correct the hyphens as well.

DenisNosnitsin1970:
Za- in the names of the scribes noted in the bottom margin, in a very typical way, is usually not a part of the scribe's name but builds a possessive construction with his name, i.e. "of XY". This is an important feature, it was discussed in a couple of publications; it would be nice to make it somehow retrievable.

PR #185, opened 2020-07-07
Update PRS5721JohnCli.xml

MarcinKrawczuk:


PRS5001-6000/PRS5721JohnCli.xml
```diff
@@ -44,14 +45,22 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
+            <person sex="1" sameAs="wd:Q317072">
+               <persName xml:lang="en" xml:id="n1">John Climacus</persName>
+              <persName xml:lang="gez" xml:id="n2">ዮሐንስ፡ በዐለ፡ መዓርግ፡</persName>
+                <persName xml:lang="grc" xml:id="n3">Ἰωάννης τῆς Κλίμακος</persName>
+               <occupation type="ecclesiastic">writer</occupation>
+               <faith type="Greek"/>
+               <birth notBefore="0525" notAfter="0579"/>
+               <floruit>He was a monk at the monastery of St. Catherine on <placeName ref="LOC5735Sinai">Mount Sinai</placeName>. His main literary work is Scala
+               Paradisi. In Ethiopian tradition he is known from quotations in <title ref="LIT4838CollectioM">Collectio Monastica</title> and
+              <title ref="LIT1933Mashaf">Maṣḥafa ḥāwi</title>.</floruit>
```
DariaElagina, line 57: I think that the correct way to point to a specific work in this case would be like that: <ref type="work" corresp="LIT4838CollectioM"></ref> for both cases

PietroLiuzzo: https://betamasaheft.eu/Guidelines/?id=references

PR #198, opened 2020-07-27
Scribes for bl orient777

DariaElagina:


thea-m:
It seems like Solomon had already created these (though with the famous za-). If you go on "Show attestations" in https://betamasaheft.eu/manuscripts/BLorient777/main, they will be listed there. In any case, I believe the correct workflow would be to merge this PR and then make a new PR to delete the records (or delete the old ones). I will approve the PR so that you can go on merging and deleting however you prefer.

DariaElagina:
Thank you, @thea-m !

PR #203, opened 2020-07-29
More persons BDLaethb5

thea-m:
PRS13107GabraKr was created erroneously and will be deleted immediately after merging (but has to be merged due to the new workflow for deleting records https://betamasaheft.eu/Guidelines/?id=create)

### DIFF
--- a/new/PRS13101ZamanfasQ.xml
+++ b/new/PRS13101ZamanfasQ.xml
@@ -1,0 +1,58 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="PRS13101ZamanfasQ" xml:lang="en" type="pers">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title>Zamanfas Qǝddus</title>
+                <editor role="generalEditor" key="AB"/>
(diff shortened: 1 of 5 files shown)


new/PRS13101ZamanfasQ.xml
```diff
@@ -0,0 +1,58 @@
...
+            <change who="DR" when="2020-07-21">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ዘመንፈስ፡ ቅዱስ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Zamanfas Qǝddus</persName>
+                    <floruit notBefore="1711" notAfter="1716"/>
```
MassimoVilla, line 50: This is perhaps a silly remark, but doesn't this mean that Zamanfas Qǝddus was born in/after 1711 and died in/before 1716?

new/PRS13115SelaKr.xml
```diff
@@ -0,0 +1,58 @@
...
+            <change who="DR" when="2020-07-29">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ሥዕለ፡ ክርስቶስ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Śǝʿla Krǝstos</persName>
+                    <floruit notBefore="1711" notAfter="1716"/>
```
MassimoVilla, line 50: The same remark as above.

thea-m:
This is the production date of the manuscript, so the only period in which we can arguably place them. The TEI definition of the floruit element is "contains information about a person's period of activity", I thought that it could be used that way. What do you think?

SusanneHummel (approved):

thea-m:
@MassimoVilla can this be merged?

MassimoVilla (approved):

PR #204, opened 2020-08-01
Create PRS13119FequraIyasus.xml

DenisNosnitsin1970:
New

### DIFF
--- a/new/PRS13119FequraIyasus.xml
+++ b/new/PRS13119FequraIyasus.xml
@@ -5,7 +5,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
     <teiHeader>
         <fileDesc>
             <titleStmt>
-                <title>Fequra Iyasus</title>
+                <title>Fəqura ʾIyasus</title>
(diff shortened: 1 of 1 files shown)


new/PRS13119FequraIyasus.xml
```diff
@@ -0,0 +1,60 @@
...
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ፍቁረ፡ ኢየሱስ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Fəqura ʾIyasus</persName>
+                    <faith type="EOTC"/>
+                    <occupation type="academic"/>
```
thea-m, line 51: academic because he is a scribe? I am not principally oposed, but I think usually we used ecclesiastic for scribes? (Or nothing at all)

DenisNosnitsin1970: OK.

PR #215, opened 2020-09-04
Added scribe for BAVet116

smaugustine:


new/PRS13142SarsaMadhen.xml
```diff
@@ -0,0 +1,63 @@
...
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ሠርጸ፡ መድኅን፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Śarṣa Madḫǝn</persName>
+                    <faith type="EOTC"/>
+                    <occupation type="other"/>
+                    <nationality type="Ethiopia"/>
+                    <birth/>
+                    <death/>
+                    <floruit/>
```
eu-genia, line 55: it would be helpful for quick disambiguation (we probably have many persons by the same name) if approximate activity time could be added (in the case of the 19th century - 1800-1900 etc.)

new/PRS13142SarsaMadhen.xml
```diff
@@ -0,0 +1,63 @@
...
+                </person>
+                <listRelation>
+                    <relation name="saws:isCopierOf" active="PRS13142SarsaMadhen" passive="BAVet116"/>
+                </listRelation>
+            </listPerson><!---->
```
eu-genia, line 60: please suppress empty elements (comments)

new/PRS13142SarsaMadhen.xml
```diff
@@ -0,0 +1,63 @@
...
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ሠርጸ፡ መድኅን፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Śarṣa Madḫǝn</persName>
+                    <faith type="EOTC"/>
+                    <occupation type="other"/>
```
eu-genia, line 51: which is "other"? if you don't know better delete the element, or specify which

new/PRS13142SarsaMadhen.xml
```diff
@@ -0,0 +1,63 @@
...
+                    <nationality type="Ethiopia"/>
+                    <birth/>
```
eu-genia, line 53: empty elements can be removed

eu-genia (changes requested):
just some minor points, thank you

new/PRS13142SarsaMadhen.xml
smaugustine: I would agree -- these were in the template and I assumed there was a reason for them.

new/PRS13142SarsaMadhen.xml
smaugustine: This is an option in the entry generator and was also in similar records, but if it's better to remove it then okay.

eu-genia (approved):

PR #216, opened 2020-09-04
Update Gitignore and delete DS_Store

smaugustine:
Invisible .DS_Store files (created by Macs) should be added to .gitignore to prevent them being added to the repo. (This is already the case in the Works repo).


PietroLiuzzo:
ok

PR #239, opened 2020-10-14
Persons for BGV-003

SusanneHummel:
Here are two new Person entries from a Ms belonging to the veroli collection.

### DIFF
--- a/new/PRS13178Yohannes.xml
+++ b/new/PRS13178Yohannes.xml
@@ -12,12 +12,12 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
             </titleStmt>
             <publicationStmt>
                 <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
-                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
-                            Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine
+                    multimediale Forschungsumgebung / Beta maṣāḥǝft</publisher>
                 <pubPlace>Hamburg</pubPlace>
                 <availability>
                     <licence target="http://creativecommons.org/licenses/by-sa/4.0/"> This file is
-                                licensed under the Creative Commons Attribution-ShareAlike 4.0. </licence>
+                        licensed under the Creative Commons Attribution-ShareAlike 4.0. </licence>
                 </availability>
             </publicationStmt>
             <sourceDesc>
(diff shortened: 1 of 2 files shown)


new/PRS13180GabraSellase.xml
```diff
@@ -0,0 +1,63 @@
...
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1"> ገብረ፡ ሥላሴ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Gabra
+                        Śǝllāse</persName>
+                </person>
+                <listRelation>
+                                       <relation name="saws:isCopierOf" active="PRS13180GabraSellase"
+                        passive="BGV003#a2">
```
MassimoVilla, line 54: I would simply write `BGV003` instead of `BGV003#a2` (which refers to the additional note only), and leave `BGV003#a2` in the `<desc/>` below by entering something like Gabra Śǝllāse is indicated as the scribe of the manuscript in `BGV003#a2`.

new/PRS13178Yohannes.xml
```diff
@@ -0,0 +1,57 @@
...
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ዮሐንስ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Yoḥannǝs</persName>
+                              </person>
+                <listRelation>
+                    <relation name="lawd:hasAttestation" active="PRS13178Yohannes" passive="BGV003"/>
```
MassimoVilla, line 52: We are recommended to add a `<desc/>` to specify the kind of relation, e.g. the name of Yoḥannǝs is mentioned in the colophon/supplication formulas/etc.; he might be a later owner, or something similar.

MassimoVilla (changes requested):
Thank you Sanne!

new/PRS13178Yohannes.xml
SusanneHummel: Thank you @MassimoVilla

MassimoVilla (approved):

new/PRS13180GabraSellase.xml
thea-m: Now I am confused - is he the scribe of the entire manuscript or only of a2? If he is the scribe of the entire manuscript, I would suggest to reformulate, if he is only the scribe of a2, I would keep #a2 in the main relation.

SusanneHummel: No, he is only the scribe of the Additio as stated in the "desc element. Therefore I pointed to msitem BGV003#a2 but I thought one should now point to the whole ms abd than in the desc-element to the specific item. 
Shall I change it again? If yes,  Do I keep the "desc-element"?

MassimoVilla: Ok, then it's my mistake. I unduly believed that he was the scribe of the entire manuscript and that this information was declared in Additio 2. I did not check the pictures, sorry. If he only wrote Additio 2, then @SusanneHummel's former statement was correct (`passive="BGV003#a2"`). Still, I would make this info unambiguous in the Desc element and state that he wrote that Additio only.

thea-m (approved):

SusanneHummel:
Thank you. I changed it back and will merge it now.

PR #245, opened 2020-10-27
Person in mnc017

MassimoVilla:
Individuals mentioned in ms MNC-017

### DIFF
--- a/new/PRS13222SahlaSe.xml
+++ b/new/PRS13222SahlaSe.xml
@@ -1,0 +1,58 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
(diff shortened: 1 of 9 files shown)


new/PRS13222SahlaSe.xml
```diff
@@ -0,0 +1,64 @@
...
+                    <persName xml:lang="gez" xml:id="n1">ሣህለ፡ ሥላሴ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Śāhla Śǝllāse</persName>
+                    <faith type="EOTC"/>
+                    <floruit notBefore="1850" notAfter="1994"/>
+                </person>
+                <listRelation>
+                    <relation name="lawd:hasAttestation" active="PRS13222SahlaSe" passive="MNC017">
+                        <desc>
+                            His name is written in a secondary hand in the supplication formulas throughout the manuscript, after the phrase <foreign xml:lang="gez">ገብርከ፡</foreign> 
+                            'your servant'. He might have been a secondary owner or donor of the manuscript.
```
SusanneHummel, line 57: When do we use patron?

MassimoVilla: Well, since that name is written in a secondary hand, the individual at issue may have owned or donated the manuscript at some point of the Ms history, but probably not commissioned its writing.
If the name of the individual was written in the main hand, then "patron" would have well be included. What do you think?

SusanneHummel: Okay, perfect. Thank you! I was just not sure, that we use "patron" only for "commissioner".

SusanneHummel (approved):

PR #246, opened 2020-10-31
Create PRS13225Hezqeyas.xml

MarcinKrawczuk:
The manuscript was written 1708-11, hence the dates in floruit.


thea-m:
Thank you! I wonder if it would not be better to add the narrower date range of the manuscript production to floruit instead of the wider span. As I understand floruit, it does not need to contain the dates of the potential life span. What do you think?
see also my remark in the other PR about the empty elements.

thea-m (approved):

PR #255, opened 2020-11-09
Update PRS5349Hippolyt.xml

MarcinKrawczuk:


PRS5001-6000/PRS5349Hippolyt.xml
```diff
@@ -46,11 +46,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
       <body>
          <listPerson>
             <person sex="1" sameAs="wd:Q207113">
-               <persName xml:lang="en" xml:id="n1">Hippolytus of Rome</persName>
+               <persName type="main" xml:lang="en" xml:id="n1">Hippolytus of Rome</persName>
                <persName xml:lang="gez" xml:id="n2">ሄጶሊጦስ፡</persName>
-               <persName xml:lang="gez" corresp="#n1">Heṗoliṭos</persName>
+               <persName type="normalized" xml:lang="gez" corresp="#n2">Heṗoliṭos</persName>
```
eu-genia, line 51: I would still say that Hippolytus, Hepolitos, Ἱππόλυτος is one and the same name in different languages, linked by corresp, not different names, and we start from Ethiopic, since this is the tradition in focus - which is n1 and the other corresp, with English highlighted as main
only abba Litos can be n2

eu-genia (changes requested):

eu-genia (approved):

PR #269, opened 2020-11-20
Update PRS2033AriusA.xml

eu-genia:


PRS2001-3000/PRS2033AriusA.xml
```diff
@@ -50,9 +51,9 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
              
                <persName xml:lang="gez" xml:id="n1" type="birth">አርዮስ፡</persName>
                <persName xml:lang="gez" corresp="#n1" type="normalized">ʾAryos</persName>
+               <persName xml:lang="en" corresp="#n1" type="main">Arius</persName>
```
thea-m, line 54: I know I have also encoded in the same often, but now I wonder: Might it be better to have  `<persName xml:lang="en" xml:id="n2" type="main">Arius</persName>`? The English and Ethiopic are versions of the same name, but actually the English version is not a translation of the Ethiopic one

thea-m (approved):

PRS2001-3000/PRS2033AriusA.xml
eu-genia: It is not a translation, but a direct correspondence, both being translations of the same Greek name. of course the correct way would go from Greek, but we do start from Ethiopic as this is our main language. I still would use the corresp, and n2 only for a really alternative name / epithet / etc.

PR #308, opened 2020-12-14
Owner of bl add24989

DariaElagina:


new/PRS13308Estifanos.xml
```diff
@@ -0,0 +1,60 @@
...
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">እስጢፋኖስ፡</persName>
+                    <persName xml:lang="gez" corresp="#n1" type="normalized">ʾƎsṭifānos</persName>
+                    <faith type="EOTC"/>
+                    <nationality type="Ethiopia"/>
+                    <floruit notBefore="1800" notAfter="1900"/>
```
thea-m, line 52: Related to the question in the manuscript record :) If there is an internal date in the manuscript, I would add it here. Maybe that is something we should discuss with the others, what do you think?

thea-m (changes requested):
similar question :)

new/PRS13308Estifanos.xml
DariaElagina: I see, I think you are right. We have at least one date documented for this person, so it would be good to add it to the record. But I wonder how: basically this could be added in a form of a free texts: This person owned a manuscript (with a ref) dated to (internal date from a manuscript). Or do you have another idea?

thea-m: I think that I have in the past taken the manuscript date in floruit without comment, but that is of course not very transparent. But with this description and a relation to the manuscript it would be already better. Could you maybe open an issue?

thea-m:
The issue, while interesting and constructive, has not answered this question at all for me. Obviously this manuscript is fine and can be merged anyway.

thea-m (approved):

PR #310, opened 2020-12-16
1st commit

MarcinKrawczuk:
As usual, floruit is the same as the approximate date of the ms.


thea-m (approved):
Thank you! (I just deleted the empty elements)

PR #317, opened 2021-01-08
updated person records related to MNC-007 and MNC-008

SusanneHummel:
Dear both,
these are updated records of persons mentioned in Mss MNC-001 and MNC-007.
Best
Sanne


MassimoVilla (approved):
Very good, thanks!

SusanneHummel:
I will merge these two records as there is nothing special

PR #323, opened 2021-01-26
Prs1277 abraham corr mk

MarcinKrawczuk:
I've read the text and it doesn't containt any historical details except the town from which he came so I haven't provided date etc.


PRS1001-2000/PRS1277Abraham.xml
```diff
@@ -32,10 +32,13 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
       <profileDesc>
-         <langUsage><language ident="en">English</language></langUsage>
+         <langUsage>
+           <language ident="en">English</language>
+           <language ident="gez">Gǝʿǝz</language>
+         </langUsage>
          <textClass>
             <keywords>
-               <term key="saint"/>
+               <term key="martyr"/>
```
thea-m, line 41: I would maybe keep saint, as they are not mutually exclusive. What do you think?

thea-m (changes requested):

thea-m (approved):

PR #353, opened 2021-04-08
Prs2659 benjamin corr mk

MarcinKrawczuk:
I'm sure you're a better expert in Christian Egypt than me - do you think that "Benjamin of Alexandria" can mean "Benjamin II"? I doubt that.


thea-m (approved):
Thank you! I agree that Benjamin II would be unlikely.

PR #377, opened 2021-06-18
Prs5585 isaiaht corr mk

MarcinKrawczuk:
It doesn't display correctly in the app,  I think it might be because /t/ was used instead of /n/ in <persName>, added a few things along the way.

### DIFF
--- a/PRS5001-6000/PRS5585Isaiaht.xml
+++ b/PRS5001-6000/PRS5585Isaiaht.xml
@@ -48,7 +48,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          <listPerson>
             <person sex="1">
                <persName xml:lang="en" xml:id="n1">Isaiah the abbot</persName>
-               <persName xml:lang="en" xml:id="n2">Isaiah of Schetis</persName>
+               <persName xml:lang="en" xml:id="n2">Isaiah of Scetis</persName>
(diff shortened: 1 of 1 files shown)


PRS5001-6000/PRS5585Isaiaht.xml
```diff
@@ -40,20 +40,24 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
       <body>
          <listPerson>
             <person sex="1">
-               <persName xml:lang="en" xml:id="t1">
-                  Isaiah the abbot
-               </persName>
-               <persName xml:lang="en" xml:id="t2">Isaiah of Schetis</persName>
-               <persName xml:lang="en" xml:id="t3">Isaiah the Hermit</persName>
+               <persName xml:lang="en" xml:id="n1">Isaiah the abbot</persName>
+               <persName xml:lang="en" xml:id="n2">Isaiah of Schetis</persName>
```
thea-m, line 51: I would propose the more common Scetis

thea-m (changes requested):
Thank you! (If no one is assigned, no one receives an email and feels responsible, but still, sorry that this nice PR was lost!)

PRS5001-6000/PRS5585Isaiaht.xml
MarcinKrawczuk: Sure

thea-m (approved):

PR #391, opened 2021-07-01
Create PRS13448Gwaggul.xml

thea-m:


new/PRS13448Gwaggul.xml
```diff
@@ -0,0 +1,58 @@
...
+            <change who="DR" when="2021-06-30">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ጓጉል፡ </persName>
+                    <persName xml:lang="gez" corresp="#n1" type="normalized">Gʷāggul </persName>
+                    <persName xml:lang="gez" corresp="#n1" type="normalized">Gʷāgul </persName>
```
DenisNosnitsin1970, line 50: I think Gwagul is not possible, because it is a (colloquial?) variant of the standard Gwangul, with "n" assimilated.

DenisNosnitsin1970 (changes requested):

new/PRS13448Gwaggul.xml
thea-m: Good, that was also my suspicion

PR #399, opened 2021-07-13
Create PRS13456Nurit.xml

thea-m:


new/PRS13456Nurit.xml
```diff
@@ -0,0 +1,57 @@
...
+        </profileDesc>
+        <revisionDesc>
+            <change who="DR" when="2021-07-13">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="2">
+                    <persName xml:lang="gez" xml:id="n1">ኑሪት፡</persName>
```
DenisNosnitsin1970, line 48: She was/is a Muslim, most probably.

DenisNosnitsin1970 (changes requested):

new/PRS13456Nurit.xml
thea-m: I agree, but the scroll has Christian content, so I didn't decide for any faith

DenisNosnitsin1970:
I think that it is not so rigid with magic texts, and it is said that elements of the magic are supra-ethnic and supra-confessional, but somehow tangible proofs were missing. In other words, I would not be amaized if an Ethiopian Muslim uses a Christian amulet, and vice versa. I think It would be useful, to indicate, at least hypothetically, that she was a Muslim.

thea-m:
Yes, there are apparently several cases, also described by Veronika Six, see also https://betamasaheft.eu/manuscripts/EMIPms00159/main (it would be interesting if in this latter case, both names occur in the same stratum and refer to the same person or whether one was added later)

PR #400, opened 2021-07-13
Update PRS11903Zakrestos.xml

MarcinKrawczuk:
There is obviously a mistake because the record was linking to itself. I am not so sure about the date 1663, it doesn't appear in Cerulli's article in this context but I'm no expert on history so I left it and just added the name of successor.


thea-m (approved):
Thank you! But it would be good to add a change element so that the public can appreciate your efforts :)

thea-m (approved):

PR #406, opened 2021-07-17
Prs13463 robel

MarcinKrawczuk:
Floruit date from the mss. cf. Wright p. 170a


new/PRS13463Robel.xml
```diff
@@ -0,0 +1,62 @@
...
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ሮቤል፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Robel</persName>
+                    <faith type="EOTC"/>
+                    <nationality type="Ethiopia"/>
+                    <birth/>
+                    <death/>
+                    <floruit when="1673"/>
```
DariaElagina, line 54: I would be in favor of defining the time of floruit in a broader way. The person was definitely active for more than one year. The date of a manuscript is just one single piece of evidence. Basically, the copying of the manuscript might have happened at the beginning, in the middle, or at the end of the scribe's career.

MarcinKrawczuk: My understanding was that floruit = yearly attestation if we have it OR conjectural range of dates if we don't. I'll keep it in mind.

DariaElagina (approved):

PR #452, opened 2021-11-01
new records

thea-m:
This is the note in which these persons are mentioned:
image

I don't understand everything, could you have a look whether I identified the personal names correctly? I also wonder whether ʾƎngədā (and Kāsā?) are Beta ʾƎsrāʾel?


DenisNosnitsin1970:
Indeed, it looks like ፈላሻው refers to qes ʾƎngədā and qes Kāsā. Not well understandable and transparent. If you want that I would distinguish the names according to my understanding, could you send me the passage in a word file?

thea-m:
created the new records

thea-m:
Can this be merged?

PR #454, opened 2021-11-02
completed

DariaElagina:
Floruit is problematic.

### DIFF
--- a/new/PRS13601BedaG.xml
+++ b/new/PRS13601BedaG.xml
@@ -5,7 +5,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
     <teiHeader>
         <fileDesc>
             <titleStmt>
-                <title>Bǝḍʿā Giyorgis</title>
+                <title>Bǝḍā Giyorgis</title>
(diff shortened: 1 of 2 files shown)


new/PRS13601BedaG.xml
```diff
@@ -0,0 +1,59 @@
...
+        </profileDesc>
+        <revisionDesc>
+            <t:change xmlns:t="http://www.tei-c.org/ns/1.0" who="DE" when="2021-11-02">Created entity</t:change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ብፃ፡ ጊዮርጊስ፡</persName>
```
DenisNosnitsin1970, line 48: Ethiopic is not the same as the transliteration.

DenisNosnitsin1970 (changes requested):

new/PRS13601BedaG.xml
DariaElagina: corrected

DenisNosnitsin1970 (approved):

PR #470, opened 2021-12-10
persons BLadd224186

thea-m:
The name Sirāk is so rare, as is the naming of illustrators, that I believe it to be highly unlikely that there are three Sirāks adding drawings to manuscripts


PRS11001-12000/PRS11365Sirak.xml
```diff
@@ -47,17 +48,16 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
             <person sex="1">
                <persName xml:lang="gez" xml:id="n1" type="birth">ሲራክ ፡</persName>
```
DenisNosnitsin1970, line 49: Please check spaces around Sirak.

DenisNosnitsin1970 (changes requested):

PR #497, opened 2022-01-28
new record

MassimoVilla:
This new record has been created manually. I know it is risky and definitely not advisable, but I have to confess that in these days BM seems slower than in the past and very often crashes.


thea-m (approved):
Thank you! Sorry for the problems - could you tell me what was the problem? Was it just slow or were there error messages?

MassimoVilla:
Thank you Thea! I do not know whether I am the only one finding some problems in loading pages, perhaps it is a problem concerning my web connection. Lately the website has become unfortunately slow and the creation of a new record failed repeatedly due to a Gateway timeout error, so I was forced to create the new record manually. Also the search function sometimes fails due to a timeout error.

thea-m:
That is annoying :( I haven't experienced this, so let's "hope" it's your connection and not a general problem. Please let me know if this persists!

PR #505, opened 2022-02-03
Create PRS13687WaldaA.xml

thea-m:
This was a nice discovery!


MassimoVilla (approved):

MassimoVilla:
Yes, this was a nice surprise, which I am sure would have passed unnoticed without the BM database

PR #508, opened 2022-02-10
Create PRS13689GabraD.xml

thea-m:
Sorry for all the individual PRs - I will try to get better at grouping them in the future.

### DIFF
--- a/new/PRS13689GabraD.xml
+++ b/new/PRS13689GabraD.xml
@@ -1,0 +1,57 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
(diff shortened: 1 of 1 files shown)


new/PRS13689GabraD.xml
```diff
@@ -0,0 +1,57 @@
...
+            <change who="DR" when="2022-02-10">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ገብረ፡ ድንግል፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Gabra Dǝngǝl</persName>
+                    <persName xml:lang="gez" xml:id="n2">ግብረ፡ ድንግል፡</persName>
```
MassimoVilla, line 50: This seems to be a scribal error. Is this spelling attested in a specific place of a manuscript?

thea-m: Yes, in fact, it's the only attested spelling, the other one is my normalisation. Do you have a suggestion how to better encode this?

MassimoVilla: There are no cases similar to this coming to my mind. I would perhaps add a note explaining that ግብረ፡ ድንግል፡ is the only attested spelling in MS g. 34, fol. 74r, but that ገብረ፡ ድንግል፡ is the normalised and presumably original form of the name.

PR #511, opened 2022-02-17
new records

thea-m:
I'm quite unsure regarding the Sofonyas situation - it would also be legitimate to create two records for Walda Iyasus and Gabra Iyasus. What do you think?


MassimoVilla:
I think that, just to be thorough, two additional records might be created for Walda Iyasus and Gabra Iyasus, leaving open the possibility that the three names refer to one single person (especially in view of ገብርከ).
No record for Kidāna Māyā, which is a scribal error.

thea-m:
Thank you! I created them.

MassimoVilla (approved):

PR #527, opened 2022-02-28
Create PRS13744Giyorgis.xml

DenisNosnitsin1970:
A new record. The scribe who copied Abb 115.


eu-genia (approved):

new/PRS13744Giyorgis.xml
```diff
@@ -0,0 +1,61 @@
...
+            <change who="DN" when="2022-02-28">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ጊዮርጊስ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Giyorgis</persName>
+                    <faith type="Orthodox"/>
```
thea-m, line 50: Sorry for intruding in a merged record - I'm just wondering, due to the possible Beta Esrael connection of the concerned manuscript, whether there is evidence on the religion of the scribe?

DenisNosnitsin1970: I would say, I do not see any immediate hints to that, at first glance, only the first text is "possibly"  a BE text.

PR #534, opened 2022-03-09
Create PRS13770EntonyosWaldaAbadia.xml

DenisNosnitsin1970:
New record.

### DIFF
--- a/new/PRS13770EntonyosWaldaAbadia.xml
+++ b/new/PRS13770EntonyosWaldaAbadia.xml
@@ -1,0 +1,62 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng"
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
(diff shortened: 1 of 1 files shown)


thea-m (commented):
Sorry, I have to ask - he is not identical to Antoine d'Abbadie? :)

new/PRS13770EntonyosWaldaAbadia.xml
```diff
@@ -55,6 +58,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                 </person>
                 <listRelation>
                     <relation name="saws:hasOwned" active="PRS13770EntonyosWaldaAbadia" passive="BNFabb122"/>
+                    <relation name="saws:isRelatedTo" active="PRS13770EntonyosWaldaAbadia" passive="PRS1071dAbbadi" cert="medium"/>
```
eu-genia, line 61: saws relations are for texts
possibly `<relation name="syriaca:possibly-identical"></relation>`?
(to me it seems relatively certain)

DenisNosnitsin1970: Yes, but theoretically an Ethiopian name like this is possible. It will be certain when I come across his Ethiopic writign and it will be the same. It will happen sooner or later.

eu-genia:
Maybe @anaiswion would be able to help if she has a minute. Possible?

DenisNosnitsin1970:
Now not necessary, I can look up in her publication.

eu-genia:
@thea-m is it ok for now with the relation?

thea-m (approved):

PR #546, opened 2022-04-19
BLadd20740

thea-m:


new/PRS13783Fasiladas.xml
```diff
@@ -0,0 +1,58 @@
...
+            <change who="DR" when="2022-04-16">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ፋሲለደስ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Fāsiladas</persName>
+                    <floruit notBefore="1600" notAfter="1799"/>
```
eu-genia, line 50: 17th-18th century? or should be only 17th as below?

eu-genia (approved):
just one small question, otherwise ok to merge )

new/PRS13783Fasiladas.xml
thea-m: It's correct (at least, it correctly renders Wright :) ), they are attested in differently dated codicological units

PR #585, opened 2022-08-01
persons BDLaethe32

thea-m:
It's possible that after reviewing, you receive an automatic email from GitHub reporting technical errors: this has to do with the current restructuring process of the app, and not with actual errors in these files. It can be ignored.


MassimoVilla (approved):
Thank you Thea. Nothing to comment!

PR #586, opened 2022-08-01
Create PRS13850WalattaS.xml

thea-m:


new/PRS13850WalattaS.xml
```diff
@@ -0,0 +1,63 @@
...
+        <body>
+            <listPerson>
+                <person sex="2">
+                    <persName xml:lang="gez" xml:id="n1">ወለተ፡ ሰንበት፡ ትበርህ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Walatta Sanbat Tǝbarrǝh</persName>
+                    <persName xml:lang="gez" xml:id="n2">ወለተ፡ ሰንበት፡ ትበርሂ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n2">Walatta Sanbat Tǝbarhi</persName>
+                    <persName xml:lang="gez" xml:id="n3">ወለተ፡ ሰንበት፡ ትበርሄ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n3">Walatta Sanbat Tǝbarhe</persName>
+                    <note>The last part of the name is attested throughout the manuscript with spelling variants.</note>
```
karljonaskarlsson, line 54: It appears that variants of ወለተ፡ ሰንበት፡ ወለተ፡ ትበርሂ፡ (with two times ወለተ፡) are also attested a couple of times.

karljonaskarlsson, line 54: As in another case previously, to me it seems reasonable to mention in which manuscript the various spellings are attested (although, as of yet, the person is of course only attested in one single manuscript... but we don't know what will come).

new/PRS13850WalattaS.xml
```diff
@@ -0,0 +1,63 @@
...
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="2">
+                    <persName xml:lang="gez" xml:id="n1">ወለተ፡ ሰንበት፡ ትበርህ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Walatta Sanbat Tǝbarrǝh</persName>
+                    <persName xml:lang="gez" xml:id="n2">ወለተ፡ ሰንበት፡ ትበርሂ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n2">Walatta Sanbat Tǝbarhi</persName>
```
karljonaskarlsson, line 51: The transcription is tricky, at least for me. Why not Tǝbarrǝhi, if we read Tǝbarrǝh as a verb form? But I am not sure about anything, so my question would be: is there a posibility to mark that the transcription is uncertain?

karljonaskarlsson (changes requested):

new/PRS13850WalattaS.xml
thea-m: I think it is ok, it will be linked to the manuscript :)

new/PRS13850WalattaS.xml
thea-m: Not sure now whether I had asked someone back then. I added a remark now.

karljonaskarlsson (approved):

PR #589, opened 2022-08-03
BDLaethg41

thea-m:


new/PRS13858MaryamS.xml
```diff
@@ -0,0 +1,58 @@
...
+        <revisionDesc>
+            <change who="DR" when="2022-08-02">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="2">
+                    <persName xml:lang="gez" xml:id="n1">ማርያም፡ ስና፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Māryām Sǝnā</persName>
```
MassimoVilla, line 49: Śǝnā in <title> (which is the etymologically correct form). How do we deal with names with fluctuation in the spelling, e.g. S/Ś?

MassimoVilla (changes requested):

new/PRS13858MaryamS.xml
thea-m: I would follow the manuscript (so S in this case, Ś was my mistake), and make a `<note>` if there is variation of spelling in the manuscript.

MassimoVilla (approved):

PR #593, opened 2022-08-04
updates

thea-m:


PRS1001-2000/PRS1016AbarraG.xml
```diff
@@ -52,7 +52,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                <persName type="alt">ʾAbbā Sǝbbār</persName>
                <nationality type="Ethiopia post-Aksumite">Christian Ethiopia</nationality>
                <faith type="EOTC">EOTC</faith>
-               <occupation>daǧǧ</occupation>
+               <persName><roleName type="title">daǧǧāzmāč</roleName></persName>
```
eu-genia, line 55: i would say occupation type military for dajjazmach actually...

PRS1001-2000/PRS1017AbarraK.xml
```diff
@@ -51,7 +51,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
                </persName>
                <nationality type="Ethiopia post-Aksumite">Christian Ethiopia</nationality>
                <faith type="EOTC">EOTC</faith>
-               <occupation>däǧǧ. </occupation>
```
eu-genia, line 54: `<occupation type="military">daǧǧāzmāč</occupation>`?

eu-genia (changes requested):

PRS1001-2000/PRS1016AbarraG.xml
thea-m: I don't know. I don't know the history of these titles well enough. I had searched for daǧǧāzmāč and found that it was most often encoded as title (but occupation type military is also used).

eu-genia: it is both, it is a title and a military function.

thea-m:
There will also be a problem with `<occupation>sch.</occupation>`, which is the source of many invalid records. I would tend to batch replace it with type academic, but probably that would not be the best choice for each single record. But I'm not willing to go through the records one by one and research each person. What would you suggest?

eu-genia:
I am perfectly fine with batch replacing to academic.

PR #632, opened 2022-09-08
fix: roleName elements with type

line-o:
It turned out that the schema requires the roleName element to have an attribute type.
After adding those both remaining files now validate.
Alternatively you could choose to change the requirement in the schema itself (in tei-betamesaheft.rng L14740 ff).
That is up to you.

### DIFF
--- a/PRS1001-2000/PRS1950Annesley.xml
+++ b/PRS1001-2000/PRS1950Annesley.xml
@@ -43,13 +43,13 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
          <listPerson>
             <person sex="1" sameAs="wd:Q5536270">
                <persName xml:id="n1">
-                   <forename>George</forename>
+                  <forename>George</forename>
                   <surname>Annesley</surname>
                </persName>
                <persName type="alt" xml:id="n2">
-                  <roleName>Viscount</roleName>Valentia</persName>
+                  <roleName type="title">Viscount</roleName> Valentia</persName>
                <persName type="alt" xml:id="n3">
-                  <roleName>Lord</roleName> Mountnorris</persName>
+                  <roleName type="title">Lord</roleName> Mountnorris</persName>
                <occupation type="other">traveler</occupation>
             </person>
          </listPerson>
--- a/PRS12001-13000/PRS12827Kasa.xml
+++ b/PRS12001-13000/PRS12827Kasa.xml
@@ -39,14 +39,14 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
             <listPerson>
                 <person sex="1">
                     <persName xml:lang="gez" xml:id="n1">ካሣ፡</persName>
-                    <persName xml:lang="gez" type="normalized" corresp="#n1"><roleName>grā getā</roleName> Kāśā</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1"><roleName type="title">grā getā</roleName> Kāśā</persName>
                     <faith type="EOTC"/>
                     <floruit notBefore="1850" notAfter="1900"/>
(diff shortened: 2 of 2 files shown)


duncdrum (approved):
nice, fixes an actual validation error, revealed by the PR ci script

eu-genia (approved):

eu-genia:
Thank you! I thought I had relaxed all I could ) (and for some reason the Oxygen did not complain)

duncdrum:
@eu-genia yes the fact that oxygen did not complain is something we should look into during out next meeting. From what I have seen so far, the validation errors revealed on CI are actual validation errors. 

Actually using different validation engines (xmllint and Oxygen-jing) is a bonus. So the feature is working as intended. We are certainly seeing things now that would have remained hidden and bitten us later. 

In some cases we need to exclude fragment files (or turn them into valid tei and adjust the xinclude accordingly).  I ll prepare something for our next meeting.

eu-genia:
can we exclude MSS with filenames containing "edition", "text", "sdc" or "facsimile" from the test? (these are all inner parts)

line-o:
Yes we can

line-o:
But they do validate

eu-genia:
that would help, thank you!
For the Frankfurt Mss the only issue I can see is the `graphic url` of the type `https://dfg-viewer.de/show/?set[mets]=http%3A%2F%2Fsammlungen.ub.uni-frankfurt.de%2Foai%2F%3Fverb%3DGetRecord%26metadataPrefix%3Dmets%26identifier%3D5421740`  but the URI is correct so I am at a loss - is there anything else that I am missing?

```
./Frankfurt/FSUor17.xml fails to validate
./Frankfurt/FSUor38.xml fails to validate
./Frankfurt/FSUor18.xml fails to validate
./Frankfurt/FSUor10.xml fails to validate
./Frankfurt/FSUor16.xml fails to validate
./Frankfurt/FSUor40.xml fails to validate
./Frankfurt/FSUor41.xml fails to validate
./Frankfurt/FSUor15.xml fails to validate
./Frankfurt/FSUor112.xml fails to validate
./Frankfurt/FSUor39.xml fails to validate
./Frankfurt/FSUor42.xml fails to validate
./Frankfurt/FSUor13.xml fails to validate
./Frankfurt/FSUor14.xml fails to validate
./Frankfurt/FSUor133.xml fails to validate
./Frankfurt/FSUffhludolfII22.xml fails to validate
./Frankfurt/FSUffhludolfII21.xml fails to validate
./Frankfurt/FSUor111.xml fails to validate
./Frankfurt/FSUor11.xml fails to validate
./Frankfurt/FSUfferueppII1.xml fails to validate
./Frankfurt/FSUor134.xml fails to validate
./Frankfurt/FSUor12.xml fails to validate
```

other than this sublist and the files to ignore I think manuscripts are all ok now

eu-genia:
on roleNames, to link issues: https://github.com/BetaMasaheft/Documentation/issues/1163

for some reason in the latest push https://github.com/BetaMasaheft/Persons/actions/runs/3039537591/jobs/4894572250 `PRS1950Annesley ` which was fixed and did pass the test again seems to fail validation (so does `PRS12827Kasa`), but possibly the PR is done from a branch which is not updated? can it be?

Is it possible, that in Persons (and other repos) we only see the files listed that failed validation (as it is now in Manuscripts) in test results?

Finally, now the schema is a bit too relaxed with the open lists to give us sufficient guidance and prompts - is it possible that I somewhat restrict back the main schema closing the lists, but in CI the files are validated against the more relaxed expanded schema?

duncdrum:
@eu-genia yes if a` PR` was started against an older version of `master` you should [rebase](https://www.w3docs.com/learn-git/git-rebase.html) it, to incorporate the latest changes from `master` / `main`.

PR #635, opened 2022-09-12
Update PRS11658Archelaus.xml

MarcinKrawczuk:
I think there is sth wrong with {titleStmt} but I don't know whether I'm allowed to change it.


eu-genia:
There are conflicts because you worked from an old version of the master branch. For simplicity, I will close this PR without merging and push your changes separately.

PR #645, opened 2022-10-24
Prs13926 ehrle

MarcinKrawczuk:
I have never done a PRS record for a European, so I hope it's correct. Also, there is no "Austria" on the list of countries available.


thea-m (approved):
Thank you! For the future, you can have a look at how I updated the persName

PR #659, opened 2022-11-28
BDLaethb3

thea-m:


new/PRS13900WaldaD.xml
```diff
@@ -0,0 +1,58 @@
...
+            <change who="DR" when="2022-09-23">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ወልደ፡ ዳዊት፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Walda Dāwit</persName>
+                    <floruit notBefore="1743" notAfter="1767"/>
```
MassimoVilla, line 50: I am probably asking something trivial: notBefore and notAfter means that Walda Dāwit was certainly born after 1743 and died before 1767. Is this what you meant? Or that he was certainly in existence during the 1743-1767 time frame?

new/PRS13902Mammo.xml
```diff
@@ -0,0 +1,60 @@
...
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ማሞ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Māmmo</persName>
+                    <persName><roleName type="title">liqe</roleName></persName>
+                    <floruit notAfter="1767"/>
```
MassimoVilla, line 51: notAfter or when (like the following records)?

new/PRS13908WaldaE.xml
```diff
@@ -0,0 +1,59 @@
...
+            <change who="DR" when="2022-09-23">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ወልደ፡ እግዚእ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Walda ʾƎgziʾ</persName>
+                    <persName><roleName type="title">grā getā</roleName></persName>
```
MassimoVilla, line 50: Perhaps "gerā getā"? Such a consonantal cluster seems odd (but it's also true that we have "Krestos" with the same cluster velar + r).

MassimoVilla (changes requested):
Thank you Thea! Basically I only have one question on how to use notBefore and notAfter

new/PRS13900WaldaD.xml
thea-m: I mean that he is attested in a manuscript dated to this period...So that he was certainly in existence at one point in this period, but not necessarily during the entire period.
You are right, it is a bit confusing.
The TEI guidelines for floruit have "contains information about a person's period of activity", the BM guidelines have "contains the date of the major active years of the record's subject". 
I believe that my use is the common one in Bm, but that's not a reason not to improve it - what would you suggest?

thea-m:
Sorry Massimo, these got lost among my tasks

new/PRS13900WaldaD.xml
MassimoVilla: Probably the best thing to do is to keep the current BM definition of floruit and stick to it. Since 1743-1767 are Walda Dāwit's "major active years" (unless further info comes to light), notBefore and notAfter are fine.

MassimoVilla (approved):

PR #660, opened 2022-12-04
Create PRS13948BerhanaMasqal.xml

DenisNosnitsin1970:
The commissioner of Sinai Eth. 6. What are these <t: ... extensions in some lines? 
I do not remember to have seen them before. Is it smth new or wrong?


eu-genia:
can you remove all t: prefixes? sometimes they are inserted for reasons i have not yet understood, as it does not usually happen...

eu-genia (approved):

PR #668, opened 2023-01-03
Dr 23 01 ayente

thea-m:
If you agree regarding the merging of the two Ayente records, I will update the reference in BLorient715 when this is merged. 
(I have no idea why GitHub makes it seem here like I edited PRS12759MamehreAynte to create PRS13962Lockhart - I deleted PRS12759MamehreAynte normally and created PRS13962Lockhart through the app).


MassimoVilla (approved):
Good! No observation from my side

PR #679, opened 2023-02-02
Create PRS-ID for Kirubel

CarstenHoffmannMarburg:
Created PRS-ID for Kirubel. Not sure, if he is an angel. Beginning of a Miracle of Jesus reads as follows: ተአምር፡ ዘገብረ፡ እግዚእነ፡ ወመድኃኒነ፡ ኢየሱስ፡ ክርስቶስ፡ በረከተ፡ ጸጋሁ፡
                                የሀሉ፡ ወኃብተ፡ ረድኤቱ፡ የሀ፻ሉ፡ ምስለ፡ ገብሩ፡ <persName ref="PRS7049Michael">ሚካኤል፡</persName> ለዓለመ፡ ዓለም፡ አሜን።
                                ወሶበ፡ ሰምዓት፡ ሰሎሜ፡ መወልዲት፡ ነገሮ፡ ለሕፃን፡ ሰብሐቶ፡ እንዘ፡ ትብል፡ ስብሐት፡ ለከ፡ ኦዘትነብር፡ በየማኑ፡ ለአብ፡ ስብሐት፡ ለከ፡ ኦዘይፀውርዎ፡
                                <persName ref="PRS0000">ኪሩቤል፡</persName>።

Note that Guesh has created an ID for some Kirubel (PRS13936Kirubel), who was active 1667-1706. Probably I am mislead and this is the same person?

### DIFF
--- a/new/PRS13976Maqaryos.xml
+++ b/new/PRS13976Maqaryos.xml
@@ -1,0 +1,63 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="PRS13976Maqaryos" xml:lang="en" type="pers">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title>ማቃርዮስ</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="CH"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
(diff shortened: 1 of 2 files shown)


thea-m:
Yes, I think your assumption is right and here the cherubim are meant, so there is no need to mark it up. But since the ID has now already been created, maybe it could be useful for the future to make it a record for Cherubim in general (none seems to exist yet).
Marqoryos is also shown here among the created files - I think you might have commited the record Kirubel on an old branch. Maybe it would avoid confusion to create a new branch from the master branch and move this Kirubel record there.

eu-genia:
Our Guidelines say that person records may be only created for persons. I do wonder if cherubs are persons. I would not mark it up, or if, then we need a term for authority files, not persons.

Carsten, I would also stop marking up the text, our (and in particular, your) focus is on manuscript cataloguing. If the cataloguing has been completed then it would be advisable to move to another record. When cataloguing only must create records for persons and places relevant for manuscript creation, and for the texts of course.

eu-genia:
Note also that this branch also contained a "new" record for Maqaryos, which was already processed in another branch. Please make sure that your master branch is clean before creating new branches, and that only corrections are there that you meant to be there. Thank you!

CarstenHoffmannMarburg:
What do you mean with clean? I understood, that I need to go back to master before I create a new branch for a new PRS-ID. Did I miss to do that or did I miss to do something else, what is neccessary? I think too many branches are confusing to me.

PR #680, opened 2023-02-02
Create PRS13979YosefArabi.xml

CarstenHoffmannMarburg:
If I understand this phrase correct, Yosef is supposed to be a contemporary of Jesus and Mary:
 ወሶበ፡ ተፈጸመ፡ ፵ዕለት፡ እምልደተ፡ እግዚእነ፡ ኢየሱስ፡ ክርስቶስ፡ በሥጋ፡ ፆረቶ፡ እግዝእትነ፡ ማርያም።
                                ወለዲተ፡ አምላክ፡ ለወልዳ፡ ወ<persName ref="PRS0000">ዮሴፍ፡ ዐራቢ፡</persName> ከመ፡ ያብእዎ፡ ውስተ፡ ቤተ፡ መቅደስ፡ በከመ፡ ተሠርዓ፡ በሕገ፡
                                <persName ref="PRS7181MosesM">ሙሴ፡</persName>ወአምጽኡ፡ ምስሌሁ፡ ዕጕለ፡ ርግብ፡
                                በከመ፡ ይቤ፡ ውስተ፡ ሕግ፡ ኵሉ፡ ዘይፈትሕ፡ ማሕፀነ፡ እሙ፡ ቅዱሰ፡ ይሰመይ፡ ለእግዚአብሔር፡

### DIFF
--- a/new/PRS13979YosefArabi.xml
+++ b/new/PRS13979YosefArabi.xml
@@ -1,0 +1,57 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="PRS13979YosefArabi" xml:lang="en" type="pers">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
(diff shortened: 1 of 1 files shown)


thea-m:
This probably is a typo for ፀራቢ, the carpenter. He already has an ID: PRS5764Joseph.
So I think that this record can be deleted

CarstenHoffmannMarburg:
Thea is right. It is even visible in the manuscript! I misinterpreted the letter because it was hardly legible! Sorry for that and thanks!

eu-genia:
Maybe it may be reasonable for Carsten at these early stages to first open an issue and ask whether any entity (places, persons, not only work) should be created, especially in cases of clearly non-Ethiopian items. While we can close without merging, for the app the ID has been used, it is not reasonable to create too many empty records.

PR #671, opened 2023-01-26
Create PRS13969GabraZaraSyon.xml

CarstenHoffmannMarburg:
Created new PersonID for owner in BMLacq681

PR #684, opened 2023-02-08
Ch 2023 01 26 prs gabra zara syon

CarstenHoffmannMarburg:


thea-m:
The same person record has already been merged yesterday. Did you carry out further edits in the old branch or is it an accidental PR due to too many branches? :)

eu-genia:
it looks like a doublet to https://github.com/BetaMasaheft/Persons/pull/671, I suggest we close this without merging.

For the future it may be better to have one branch for all new persons for one manuscript to avoid confusion.

And branches must always be created from master and not from earlier branches ;) Thank you!

eu-genia:
(now i see that this is the same old branch which we had actually deleted yesterday which got reopened. for the future, if you need to correct something in a record that has already been merged, create a new branch from master; never use old branches. thank you!)

CarstenHoffmannMarburg:
Yes, I got confused on the branches and pull requests! I thought it might be better to do it once to often rather than to once too less. I am sorry for that!

thea-m:
No worries, completely normal :)

PR #704, opened 2023-04-22
Pers-Names for JJ encoding

gnizla:
Some pers-names that were missing in BM records


thea-m (approved):
Thanks a lot, really nice to have these! I also added wikidata ID and BM ID to wikidata

PR #716, opened 2023-05-17
Jared (descendant of Adam)

gnizla:


thea-m:
Thank you! I committed a few minimal additions directly, otherwise the file is fine!
But it seems that you accidentally added this new file to the already deleted branch "JJ-names", which leads to conflicts and should be avoided. I copied the file to a new branch and will assign you as reviewer of the PR, so that this PR can be closed.

PR #724, opened 2023-06-13
Prs7680 os corr mk

MarcinKrawczuk:


PRS7001-8000/PRS7680Os.xml
```diff
@@ -45,19 +48,26 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?><TEI
...
          <listPerson>
             <person sex="1">
-               <persName>
-                  
-                  ʿOṣ
-               </persName>
+               <persName xml:lang="gez" xml:id="n1">ዖጽ፡</persName>
+               <persName xml:lang="gez" type="normalized" corresp="#n1">ʿOṣ</persName>
                <!-- St -->
+  <floruit>His name appears in some lists of the <ref type="pers" corresp="PRS12196NineSaints">Nine Saints</ref>, usually substituting for <ref type="pers" corresp="PRS8711sehmas">Ṣǝḥmā</ref>. The place of his origin is Kʷǝzārā, possibly a corruption
```
CarstenHoffmannMarburg, line 61: I would suggest to speak about a "transcription" rather than "corruption" - evem though it is a very odd way of transcribing Caesarea/Καισάρεια/قيسارية . Another option might be to think about الخزر Khazars, but this record is not the right place to discuss several options in parallel.

CarstenHoffmannMarburg (changes requested):

PR #725, opened 2023-06-15
Haile slassie

Addisie:


new/PRS14026Haile_Selassie.xml
```diff
@@ -0,0 +1,64 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="PRS14026Haile_Selassie" xml:lang="en" type="pers">
```
eu-genia, line 4: NB: we have to change the ID AFTER the merging, we should have no dashes in IDs, just letters
But we should first merge

PR #727, opened 2023-06-19
Persons in BDLeu6

MassimoVilla:
Dear both, 
two persons (or perhaps one? It is not clear from the MS since names have been overwritten on previously erased names) contained in MS BLDeu6.
I'm also adding a note in the MS record specifying that the two names might possibly refer to one single person.


new/PRS14037Zamaryam.xml
```diff
@@ -0,0 +1,58 @@
...
+        <revisionDesc>
+            <change who="MV" when="2023-06-19">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ዘማርያም፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">ZaMaryām</persName>
```
thea-m, line 49: I believe our transliteration system would have Zamāryām

thea-m (changes requested):
Thank you!

thea-m (approved):

PR #740, opened 2023-09-12
updates

thea-m:


smaugustine (approved):
Is diesNatalis the right encoding in these cases? These later abbots died mostly typical deaths without persecution.

PR #768, opened 2023-11-28
Created PRS for BMLacq679

CarstenHoffmannMarburg:
I created four PRS IDs for people mentioned in BMLacq679 as discussed in [#2310](https://github.com/BetaMasaheft/Documentation/issues/2310). 

I came to the conclusion that the person PRS14220HeggaYohannes is the father of PRS14221Menilek. 

I am still puzzled as to whether it is possible to establish a family relationship for the others as well. 
PRS14222TawaldaMadhen can be seen to be the heir or heiress of the former, if we accept this name in connection with አርስት (succession, inheritance, etc.). 

PRS14223Maqaryos is mentioned as ራብአዊ, but it is not clear what ራብአዊ stands for. He could be the fourth in the line of succession (e.g. the great-grandson), or the fourth in the list of blessed people. It is unlikely that he is a high-ranking dignitary such as a patriarch or abbot, as he has no title to indicate this.

The text is given twice on two successive folios. There is some uncertainty about the name Həgga (ʾƏgziʾabəḥer) Yoḥannəs, as he is mentioned once as Həgga Yoḥannəs and the second time as Həgga ʾƏgziʾabəḥer Yoḥannəs. However, I think that the encircling of ʾƏgziʾabəḥer tells us that the name ʾƏgziʾabəḥer was a mistake of the scribe.

Screenshot 2023_Maqaryos

NB. I did not understand how to reference an issue. When I press the 'reference' button in the taskbar, nothing changes except for the # that is introduced.

### DIFF
--- a/new/PRS14221Menilek.xml
+++ b/new/PRS14221Menilek.xml
@@ -5,7 +5,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
     <teiHeader>
         <fileDesc>
             <titleStmt>
-                <title>Mənilək</title>
+                <title>Mǝnilǝk</title>
                 <editor role="generalEditor" key="AB"/>
                 <editor key="CH"/>
                 <funder>Akademie der Wissenschaften in Hamburg</funder>
@@ -46,7 +46,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
             <listPerson>
                 <person sex="1">
                     <persName xml:lang="gez" xml:id="n1">ምኒልክ</persName>
-                    <persName xml:lang="gez" type="normalized" corresp="#n1">Mənilək</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Mǝnilǝk</persName>
                     <floruit notBefore="1800" notAfter="1987">19th or 20th century</floruit>
                 </person>
                 <listRelation>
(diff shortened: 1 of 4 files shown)


eu-genia:
For issues you need to copy paste the link.
NB: the capital schwa is not Ə (Unicode 018F) but Ǝ (Unicode 018E), 
Do make sure that you use the correct Unicode for ǝ as well (01DD)  AND NOT ə (Unicode 0259).

ǝ (01DD) corresponds to Ǝ (Unicode 018E) when you switch to capital in Word, try it out.

new/PRS14221Menilek.xml
```diff
@@ -0,0 +1,59 @@
...
+        </profileDesc>
+        <revisionDesc>
+            <change who="CH" when="2023-11-28">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ምኒልክ</persName>
```
eu-genia, line 48: I think we always need ፡ at the end

new/PRS14223Maqaryos.xml
```diff
@@ -0,0 +1,60 @@
...
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">መቃርዮስ</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Maqāryos</persName>
+                    <persName xml:lang="gez" type="nick" xml:id="n2">ዘክበበ፡ ገጹ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n2"></persName>
```
eu-genia, line 51: transcription?

eu-genia (changes requested):

new/PRS14221Menilek.xml
CarstenHoffmannMarburg: OK, done for these two. Alas, I think, it happened very often in my records.

eu-genia (approved):

PR #780, opened 2023-12-19
Create PRS14231TaklaM.xml

thea-m:


CarstenHoffmannMarburg (approved):

new/PRS14231TaklaM.xml
```diff
@@ -0,0 +1,58 @@
...
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ተክለ፡ ሚካኤል፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Takla Mikāʾel</persName>
+                    <floruit notBefore="1925" notAfter="1974"/>
+                </person>
+                <listRelation>
+                    <relation name="lawd:hasAttestation" active="PRS14231TaklaM" passive="FCCOSEth001"/>
```
eu-genia, line 53: not hasOwned?

thea-m: I prefer to stay more general when we do not have more information. The catalogue has "original owner", but they often write this also on persons we would rather qualify as commissioners.

eu-genia (approved):

PR #790, opened 2024-01-16
PRS in EMML1832 (no.6)

CarstenHoffmannMarburg:
I created new PRS IDs for the note on incensing in EMML1832 according to our discussion in https://github.com/BetaMasaheft/Documentation/issues/2458.

### DIFF
--- a/new/PRS14265Talqa.xml
+++ b/new/PRS14265Talqa.xml
@@ -1,0 +1,61 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="PRS14265Talqa" xml:lang="en" type="pers">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title>Ṭalqā</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="CH"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
+                            Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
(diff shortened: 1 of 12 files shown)


CarstenHoffmannMarburg:
There is an ongoing discussion on https://github.com/BetaMasaheft/Documentation/issues/2458 about whether the entities can properly be considered personal names. They probably need to be deleted and not recorded.

eu-genia (commented):
I am not sure what to do of this PR
I would not really encode any of these
The term in phrases preceded by ba'al is surely always a place
But also the phrases ba'ala medra XX, even if some of these XX may originally be persons, are referring to my understanding to plots of land, overannotating sometimes only creates confusion

CarstenHoffmannMarburg:
In this pull request, all 12 files need to be deleted because they are either attested place names or unclear but unlikely to be personal names, so we have decided not to mark them up in the record. So this pull request needs to be approved and merged, and then the records need to be deleted in a new pull request.

eu-genia:
Done
Speaking about this: this is what has been merged and synced
image

now there are only two persons between 14237 and 14286

but this is what has been created in the app and is still there, so has not been synced either way - over 30 files
image

Krǝstos Mǝkḥǝna has been created twice

some of these may still be awaiting merge but please do check if there are any records that have gotten lost on the way

CarstenHoffmannMarburg:
Records that definitely needs to be deleted are:
PRS14261Mahbar
PRS14262Zengita
PRS14263Wambaya
PRS14264Yalaela

CarstenHoffmannMarburg:
Records that are under discussion in branch PRS in EMML (no.7) are:
 <persName ref="PRS14277Mamas">ማማስ።</persName>
PRS14278AmlakMoa
PRS14279AmlakaBena
PRS14280Simeon
PRS14281Senoda
PRS14282YohannesKama
PRS14283YeheyyesTaameno
PRS14284Endreyas
PRS14285ZaraBuruk

CarstenHoffmannMarburg:
> Krǝstos Mǝkḥǝna has been created twice

This probably happened because the internet was cut off during a train journey when I was encoding this person.
PRS14248KrestosMekhena was marked up in EMML1832.

eu-genia:
if you have a blackout before you could download when creating a record do not recreate, just go to betamasaheft.eu/ID and download from there

CarstenHoffmannMarburg:
> if you have a blackout before you could download when creating a record do not recreate, just go to betamasaheft.eu/ID and download from there

I have this problem right now. I wanted to create a person (QatsalaGiyorgis), but it did not work out. Instead, I got "504 Gateway Time-out". If I type betamasaheft.eu/ID, I get "An error has occurred - something has gone wrong. Sorry! This error has been generated by the application."

eu-genia:
You should not type `ID `I meant `ID` is the placeholder for the record you created but could not download, in this case PRS14333QatsalaGiyorgis

so you should go to betamasaheft.eu/PRS14333QatsalaGiyorgis and from there to the XML or alternatively directly to https://betamasaheft.eu/PRS14333QatsalaGiyorgis.xml , copy the XML file, edit it, save as `PRS14333QatsalaGiyorgis  `in the new folder and sync as usual

PR #794, opened 2024-01-23
PRS in EMML1832 (no.9)

CarstenHoffmannMarburg:


new/PRS14302Mamas.xml
```diff
@@ -0,0 +1,60 @@
...
+        </profileDesc>
+        <revisionDesc>
+            <change who="CH" when="2024-01-23">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person>
+                    <persName xml:lang="gez" xml:id="n1"></persName>
```
thea-m, line 48: fidal missing

new/PRS14298ZaraAbreham.xml
```diff
@@ -0,0 +1,59 @@
...
+    <text>
+        <body>
+            <listPerson>
+                <person>
+                    <persName xml:lang="gez" xml:id="n1">ዘርአ፡ አብርሃም፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Zarʾa ʾAbrǝhām</persName>
+                    <floruit notBefore="1480" notAfter="1550">End of 15th century or first half of 16th century</floruit>
+                </person>
+                <listRelation>
+                    <relation name="snap:FatherOf" active="PRS14298ZaraAbreham" passive="PRS14299Faraday"/>
```
thea-m, line 53: maybe it could be noted that a spiritual relationship is likely? (if that is the case)

thea-m (changes requested):

new/PRS14298ZaraAbreham.xml
eu-genia: I think the relationship is spiritual in which case `snap `is not appropriate

CarstenHoffmannMarburg: Do you think it is appropriate to transform it to a dc:relation or would you suggest do omit this information.
E.g.
```
<relation name="dc:relation" active="PRS14298ZaraAbreham" passive="PRS14299Faraday"><desc>He is called the "father"
                        of <persName ref="PRS14299Faraday"/>, what is probably a spiritual relationship.</desc></relation>
```

thea-m: I think the dc:relation is fine

eu-genia (approved):

thea-m (approved):

PR #798, opened 2024-01-27
Prs14312 euphrasia

MarcinKrawczuk:
Could you please double-check that she is not in the database? I would imagine most saints are there but I cannot find her.


thea-m (approved):
I didn't find her either, thank you!

PR #801, opened 2024-01-30
Prs14314 maximus

MarcinKrawczuk:
I couldn't find him in the database.

Also, if I remember correctly, even if two saints always appear together we are expected two create two records. So if this one is OK, I will create the one for Domitius.


thea-m (approved):
Great, thank you

PR #816, opened 2024-02-17
Prs2470 bartalom corr mk 14.02.2024

MarcinKrawczuk:
I could add some more details perhaps but I don't have access to Getatchew's book. But I think it's enough to distuinguish him clearly from other B's.


thea-m (approved):
Great, thank you!

PR #819, opened 2024-02-21
PRS for BLorient13271

CarstenHoffmannMarburg:
I created two records for the owner and the scribe of BLorient13271.

### DIFF
--- a/new/PRS14343WaldaMaryam.xml
+++ b/new/PRS14343WaldaMaryam.xml
@@ -1,0 +1,58 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="PRS14343WaldaMaryam" xml:lang="en" type="pers">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title>Walda Māryām</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="CH"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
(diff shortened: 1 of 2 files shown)


new/PRS14343WaldaMaryam.xml
```diff
@@ -0,0 +1,58 @@
...
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ወልደ፡ ማርያም፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Walda Māryām</persName>
+                    <floruit notBefore="1800" notAfter="1899">19th century</floruit>
+                </person>
+                <listRelation>
+                    <relation name="saws:isCopierOf" active="PRS14343WaldaMaryam" passive="BLorient13271"/>
```
eu-genia, line 53: NB
there is an homonymous scribe of BLadd16255 (PRS11095WaldaM)

and many other scribes of the same name
PRS11958WaldaM
PRS11596WaldaM
PRS12667WaldaMaryam
PRS12918WaldaMaryamScribe
PRS12824WaldaMaryam
at some point the hands should be compared...

eu-genia (approved):

new/PRS14343WaldaMaryam.xml
CarstenHoffmannMarburg: Yes, I agree. For this record, I had a quick look in all records with that name, but I did not find any that is recorded for the same century and for this occupation. It could be helpful, when occupations and floruits were added in all PRS records. However for BLorient13223, we do have no images. Therefore comparison of hands is not possible at the moment. One could rather only indicate possible identifications if there are any.

new/PRS14343WaldaMaryam.xml
```diff
@@ -0,0 +1,58 @@
...
+            <change who="CH" when="2024-02-21">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ወልደ፡ ማርያም፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Walda Māryām</persName>
+                    <floruit notBefore="1800" notAfter="1899">19th century</floruit>
```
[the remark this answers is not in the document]
CarstenHoffmannMarburg: OK. I thougt this is already given in the relation "isCopierOf". To add in occupation, I assume, that
 `<occupation type="literary">scribe</occupation>` is correct?

new/PRS14343WaldaMaryam.xml
eu-genia: Yes I have checked now (updated above) they are all most probably different and do not overlap. Still let us add for all scribes `<occupation type="literary">scribe</occupation>`, then it is easier to search and filter.

eu-genia (approved):

PR #854, opened 2024-05-06
Prs14418 nob

MarcinKrawczuk:
1) The floruit is after the paleographical dating of the mss.
2) I think that there is some chance that he might be the same person as PRS13927Nob, if you agree - is there a way to show it via a relation in the person record?
Thank you!

### DIFF
--- a/new/PRS14418Nob.xml
+++ b/new/PRS14418Nob.xml
@@ -51,10 +51,11 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                     <nationality type="Ethiopia"/>
                     <birth/>
                     <death/>
-                    <floruit notBefore="1350" notAfter="1500"/>
+                    <floruit notBefore="1350" notAfter="1500">Second half of the 14th or the 15th Century</floruit>
                 </person>
                 <listRelation>
                     <relation name="lawd:hasAttestation" active="PRS14418Nob" passive="BNFabb126"/>
+                    <relation name="dc:relation" active="PRS14418Nob" passive="PRS13927Nob"><desc>Possibly the same person as <persName ref="PRS13927Nob"/>.</desc></relation>
                 </listRelation>
             </listPerson><!---->
         </body>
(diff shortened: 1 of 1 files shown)


CarstenHoffmannMarburg:
Thank you very much!

There is a way to mark a potentially identical person with dc:relation. I have a similar case in another record. I think the correct way to encode it, will be: 
```
<relation name="dc:relation" active="PRS14418Nob" passive="PRS13927Nob"><desc>Possibly 
                        the same person as <persName ref="PRS13927Nob"/>.</desc></relation>
```
Probably @eu-genia or @thea-m will know better!

CarstenHoffmannMarburg (approved):

new/PRS14418Nob.xml
```diff
@@ -0,0 +1,62 @@
...
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ኖብ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Nob</persName>
+                    <faith type="EOTC"/>
+                    <nationality type="Ethiopia"/>
+                    <birth/>
+                    <death/>
+                    <floruit notBefore="1350" notAfter="1500"/>
```
CarstenHoffmannMarburg, line 54: I think, it would be nice to add the time-span expressively, i.e.:
<floruit notBefore="1350" notAfter="1500">Second half of 14th or 15th century<floruit>

new/PRS14418Nob.xml
```diff
@@ -0,0 +1,62 @@
...
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ኖብ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Nob</persName>
+                    <faith type="EOTC"/>
+                    <nationality type="Ethiopia"/>
+                    <birth/>
+                    <death/>
```
CarstenHoffmannMarburg, line 53: It is better to delete empty spaces as <birth/> and <death/> in this case.

MarcinKrawczuk:
Thank you, I think I've made the corrections you suggested.

wt., 7 maj 2024 o 06:55 Carsten Hoffmann ***@***.***>
napisał(a):

CarstenHoffmannMarburg (approved):

PR #855, opened 2024-05-07
PRS in BLorient2083

CarstenHoffmannMarburg:
I want to create six PRS records for persons, that are attested in BLorient2083. PRS14423Malakotawit, who commissioned the manuscript might be identical with a prominent figure. However this is not certain but rather speculation. Do you agree to mark this possible identification with dc:relation or do you think this is too speculative?


new/PRS14424TasfaIyasus.xml
```diff
@@ -0,0 +1,58 @@
...
+            </langUsage>
+        </profileDesc>
+        <revisionDesc>
+            <change who="CH" when="2024-05-02">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person>
```
eu-genia, line 47: sex="1"

eu-genia (changes requested):

eu-genia (approved):

PR #874, opened 2024-06-18
PRS in BLorient8819 (owners)

CarstenHoffmannMarburg:
Unfortunately, the floruits could not be given more precisely.

### DIFF
--- a/new/PRS14479SahlaSellase.xml
+++ b/new/PRS14479SahlaSellase.xml
@@ -47,7 +47,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                 <person sex="1">
                     <persName xml:lang="gez" xml:id="n1">ሣህለ፡ ሥላሴ፡</persName>
                     <persName xml:lang="gez" type="normalized" corresp="#n1">Śāhla Śǝllāse</persName>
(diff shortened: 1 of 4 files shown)


new/PRS14479SahlaSellase.xml
```diff
@@ -0,0 +1,59 @@
...
+            <change who="CH" when="2024-06-18">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ሣህለ፡ ሥላሴ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Śāhla Śǝllāse</persName>
+                    <floruit notBefore="1600" notAfter="1900">17th/18th/19th century</floruit>
```
DenisNosnitsin1970, line 50: Here and below, I would write 17th–19th century, or indicate in words (I think this is the more common style in BM, though not the only one). I would also indicate as the second number 1899, because the year 1900 is already the next century, isn't it?

DenisNosnitsin1970 (changes requested):

DenisNosnitsin1970 (approved):

PR #892, opened 2024-07-24
PRSs in Uppsala, O Etiop 8

karljonaskarlsson:


CarstenHoffmannMarburg:
> Sorry, @CarstenHoffmannMarburg , I found some more people and thought I should add them here. I have a question: Do we generally use the "hasOwned" relation if someone's name has been secondarily added in a supplication, or is it possible to use "hasAttestation"?

I usually use 'hasOwned' when I think a manuscript has been the legitimate property of someone for some time. If someone writes their name in the manuscript, it usually means something. If it is not the scribe, it should usually be the owner or someone related to the owner. Sometimes it can be difficult to decide, but usually it is quite clear. I think more than 90% probability is good enough.

new/PRS14512Pistis.xml
```diff
@@ -0,0 +1,61 @@
...
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="2">
+                    <persName xml:lang="gez" xml:id="n1">ጲስጢስ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Ṗisṭis</persName>
+                    <faith type=""/>
+                    <birth/>
+                    <death/>
```
CarstenHoffmannMarburg, line 51: I personally remove all empty strings (i.e. <birth/>, <death/> and <faith type=""/>) because if they remain, they will attract unnecessary attention from the user.

CarstenHoffmannMarburg (approved):

new/PRS14512Pistis.xml
CarstenHoffmannMarburg: I have not come across such a case of very many community members as owners so far. What I have seen are for example protective scrolls in which an owner is named, who desired protection for himself as well as to some members of his family. In such cases, I encoded the first person as the owner and the other one or two persons as attestations. To my opinion it is helpful to encode a set of persons, which are related to each other. I think, this could be helpful to identify some of them, who are mentioned together as well in other places. However, this is also possible, when the names are listed in history without markup, as Denis suggested.

DenisNosnitsin1970: If the persons are related, sure, and this is most frequently indicated (husband, wife, and sons). Larger groups of people appear in more recent books, as the ways of book aquisition modifies: not only individuals and families, but communities commission / buy books... Then many people can be mentioned.

karljonaskarlsson: In this case, there are numerous layers of erasure in most of the supplications at the end of texts or text parts. I think I have gone through the manuscript quite carefully (although, of course, maybe I missed some), and there are some names that occur frequently, while others appear only one or two times.

Now I have created files for them, so in this case I will mention them all. I think your reasoning about community members sounds reasonable, so I will encode the main names as owners and the others as "hasAttestation".

Thank you both!

CarstenHoffmannMarburg (approved):

PR #894, opened 2024-07-25
Create PRS in BLorient11763 (owners)

CarstenHoffmannMarburg:
I did not find any information on Miss M. Gresley.

### DIFF
--- a/new/PRS14518Gresley.xml
+++ b/new/PRS14518Gresley.xml
@@ -1,0 +1,56 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="PRS14518Gresley" xml:lang="en" type="pers">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title>M. Gresley</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="CH"/>
+                <funder>Akademie der Wissenschaften in Hamburg</funder>
+            </titleStmt>
+            <publicationStmt>
+                <authority>Hiob-Ludolf-Zentrum für Äthiopistik</authority>
+                <publisher>Die Schriftkultur des christlichen Äthiopiens und Eritreas: Eine multimediale
+                            Forschungsumgebung / Beta maṣāḥǝft</publisher>
+                <pubPlace>Hamburg</pubPlace>
+                <availability>
+                    <licence target="http://creativecommons.org/licenses/by-sa/4.0/"> This file is
+                                licensed under the Creative Commons Attribution-ShareAlike 4.0. </licence>
+                </availability>
+            </publicationStmt>
+            <sourceDesc>
+                <p/>
+            </sourceDesc>
+        </fileDesc>
+        <encodingDesc>
+            <p>A digital born TEI file</p>
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p/>
+            </abstract>
+            <langUsage>
+                <language ident="en">English</language>
+            </langUsage>
+        </profileDesc>
+        <revisionDesc>
+            <change who="CH" when="2024-07-25">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="2">
+                    <persName><forename>M.</forename> <surname>Gresley</surname></persName>
+                    <floruit notBefore="1900" notAfter="1978">20th century</floruit>
+                </person>
+                <listRelation>
+                    <relation name="saws:hasOwned" active="PRS14520BakweraSeyon" passive="BLorient11763"/>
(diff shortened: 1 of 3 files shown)


new/PRS14520BakweraSeyon.xml
```diff
@@ -0,0 +1,58 @@
...
+            <change who="CH" when="2024-07-25">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person>
+                    <persName xml:lang="gez" xml:id="n1">በኵረ፡ ጽዮን፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Bakʷǝra Ṣǝyon</persName>
+                    <floruit notBefore="1600" notAfter="1944">17th-20th century</floruit>
```
DenisNosnitsin1970, line 50: Here and elsewhere: If you have a chance to encode the years very exactly in floruit, there is no reason to be less exact in the addition you write between the brackets, because this span will be visualised. In other words, you could better indicate between the brackets:  <floruit notBefore="1600" notAfter="1944">17th-20th century, but not after 1944</floruit>.

new/PRS14519WaldaSamuel.xml
```diff
@@ -0,0 +1,58 @@
...
+            <change who="CH" when="2024-07-25">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ወልደ፡ ሳሙኤል፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Walda Sāmuʾel</persName>
+                    <floruit notBefore="1600" notAfter="1944">17th-20th century</floruit>
```
DenisNosnitsin1970, line 50: 350 years! can it be reduced a little?

new/PRS14518Gresley.xml
```diff
@@ -0,0 +1,56 @@
...
+        <revisionDesc>
+            <change who="CH" when="2024-07-25">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="2">
+                    <persName><forename>M.</forename> <surname>Gresley</surname></persName>
+                    <floruit notBefore="1900" notAfter="1978">20th century</floruit>
```
DenisNosnitsin1970, line 48: Indicate it exactly: 20th century, but not after 1978 (see below).

DenisNosnitsin1970 (changes requested):

new/PRS14519WaldaSamuel.xml
CarstenHoffmannMarburg: I think, it is more likely for the 17th or 18th century, when this manuscript was supposed to be written. However, since magic scrolls often change hands and names are effaced and written over, I cannot exclude a later date. I can probably reduce to 1600 to 1899

new/PRS14518Gresley.xml
CarstenHoffmannMarburg: I think it is better to change to 1900 to 1999. We do not know, whether this was a young or an old Miss, when she bequeathed the scroll to the British Museum.

karljonaskarlsson (approved):

new/PRS14518Gresley.xml
eu-genia: BUT we have the catalogue as the terminus ante quem

new/PRS14518Gresley.xml
```diff
@@ -0,0 +1,56 @@
...
+                </person>
+                <listRelation>
+                    <relation name="saws:hasOwned" active="PRS14520BakweraSeyon" passive="BLorient11763"/>
```
eu-genia, line 51: Here the active element seems wrong

new/PRS14518Gresley.xml
eu-genia: Is it sure that it was post 1900? 
Gresley are a known family https://en.wikipedia.org/wiki/Gresley_baronets and a certain Miss M. Gresley from the family published in 1860. https://research-repository.st-andrews.ac.uk/bitstream/handle/10023/25063/Luxford_2018_ManuscriptsEandW_Ridware_AAM.pdf

new/PRS14519WaldaSamuel.xml
```diff
@@ -0,0 +1,58 @@
...
+                </person>
+                <listRelation>
+                    <relation name="saws:hasOwned" active="PRS14520BakweraSeyon" passive="BLorient11763"/>
```
eu-genia, line 53: ??

new/PRS14520BakweraSeyon.xml
```diff
@@ -0,0 +1,58 @@
...
+                </person>
+                <listRelation>
+                    <relation name="saws:hasOwned" active="PRS14520BakweraSeyon" passive="BLorient11763"/>
```
eu-genia, line 53: this is the one record where the active is correct ))

eu-genia (changes requested):

new/PRS14518Gresley.xml
CarstenHoffmannMarburg: It is possible that this is the family. However, the manuscript was given to the BL in 1944. It is unlikely that this is the same Miss M. Gresley who published in 1860. 

It is always problematic to add a floruit when little is known about a person. This person could be a very young woman when she presents the manuscript, which she may have inherited, but she could also be a rather old woman, although the title 'Miss' suggests that she was not married in that year. In her case, the floruit (= the main period of her activity) could be either mostly before 1944 or after 1944. The date of the catalogue should not be decisive here.

PR #909, opened 2024-09-04
PRS in BLorient4865

CarstenHoffmannMarburg:
I could not figure out any prominent person with the name G. Baker in around 1896.

### DIFF
--- a/new/PRS14552WKTaytu.xml
+++ b/new/PRS14552WKTaytu.xml
@@ -1,0 +1,58 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="PRS14552WKTaytu" xml:lang="en" type="pers">
+    <teiHeader>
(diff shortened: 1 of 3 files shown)


new/PRS14552WKTaytu.xml
```diff
@@ -0,0 +1,58 @@
...
+        </profileDesc>
+        <revisionDesc>
+            <change who="CH" when="2024-09-04">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="2">
+                    <persName xml:lang="gez" xml:id="n1">ወለተ፡ ኪዳን፡ ጣይጡ፡</persName>
```
karljonaskarlsson, line 48: Should it be ጣይቱ? Or should the transcription be changed?

karljonaskarlsson (changes requested):

karljonaskarlsson:
I also couldn't find G. Baker

new/PRS14553WalattaMaryam.xml
nafisa-valieva, line 1: Dear Carsten, may be you can add evidence for 1984.

eu-genia (approved):

new/PRS14553WalattaMaryam.xml
CarstenHoffmannMarburg: The reason for determining the floruit to 1894 was the acquisition of the manuscript by the British Museum in that year. However, I think it is not neccessarily the case, that people die, when their manuscripts are entering a library. I think, the floruits are in most case rather estimations and for that reason, I decided to change ist to `<floruit notBefore="1700" notAfter="1899">18th/19th century</floruit>` (both in PRS14553WalattaMaryam and PRS14552WKTaytu).

PR #913, opened 2024-09-12
Owners of BLorient12959

CarstenHoffmannMarburg:
I have found no person named J. Malcolm-Davies or ʾƎssǝta Śǝllāse.

However, when I created a record for ʾƎssǝta Śǝllāse, I made the mistake of typing EssetaSǝllāse to the ID in the form and this resulted in an error because of the letter ǝ, which is not allowed in an ID. I could not change the ID, so I created a new one, which by mistake ended up as PRS14563EssetaSellase_ (with a low line '_' at the end). I hope this low line is not a big problem. Otherwise I may have to create a new ID again. Sorry for the confusion!

### DIFF
--- a/new/PRS14563EssetaSellase_.xml
+++ b/new/PRS14563EssetaSellase_.xml
@@ -1,0 +1,58 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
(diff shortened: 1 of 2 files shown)


new/PRS14563EssetaSellase_.xml
```diff
@@ -0,0 +1,58 @@
...
+        </profileDesc>
+        <revisionDesc>
+            <change who="CH" when="2024-09-12">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="2">
+                    <persName xml:lang="gez" xml:id="n1">እስተ፡ ሥላሴ፡</persName>
```
karljonaskarlsson, line 48: I guess እሴተ፡ , ʾƎsseta , would be the expected form, and if I don't overlook anything, it seems to be what is in the catalogue

karljonaskarlsson (changes requested):

eu-genia (approved):

PR #918, opened 2024-09-19
Rename PRS14562EssetaSellase

CarstenHoffmannMarburg:
I changed the name of the record and added a change relation (formerlyAlsoListedAs). I hope, that was correct, even though it is the same number in active and passive.


eu-genia (approved):
thank you. it is correct but indeed unnecessary when the number does not change

PR #927, opened 2024-10-16
Create PRS13558Walatta.xml

MarcinKrawczuk:
The record seems OK but I don't see the file in my folder.


eu-genia (approved):
Indeed, BLorient686 points to this record but it had not been previously merged

PR #932, opened 2024-10-18
1st commit

MarcinKrawczuk:


eu-genia:
it seems to me that the records already exist on master

MarcinKrawczuk:
It does indeed, sorry. It seems that the "wife-of" relation does not
display, I cannot find it in the app.

pt., 18 paź 2024 o 22:33 Eugenia ***@***.***> napisał(a):

PR #944, opened 2024-10-24
Create PRS14534GabraMikael.xml

CarstenHoffmannMarburg:
I seen it as a "stale" branch. I probably forgot to push and pull it. Only a stub was in the app so far.


eu-genia (approved):
it is OK but you should not check whether the app has the stub you should check what is in the MASTER branch. We work with GitHub and make sure that the processed branches are merged on to MASTER. You do not need the app for working at all.

PR #951, opened 2024-11-07
PRS in bavet115

MassimoVilla:
Three individuals mentioned in BAVet115

### DIFF
--- a/new/PRS14584GabraSe.xml
+++ b/new/PRS14584GabraSe.xml
@@ -1,0 +1,62 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="PRS14584GabraSe" xml:lang="en" type="pers">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title>Gabra Śǝllāse</title>
+                <editor role="generalEditor" key="AB"/>
+                <editor key="MV"/>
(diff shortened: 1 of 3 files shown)


new/PRS14584GabraSe.xml
```diff
@@ -0,0 +1,61 @@
...
+            <change who="MV" when="2024-11-07">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ገብረ ሥላሴ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Gabra Śǝllāse</persName>
+                    <faith type="EOTC"/>
```
karljonaskarlsson, line 50: Do you have positive evidence for his faith (same for the others)?

MassimoVilla: Only the fact that they are mentioned in a Maṣḥafa gǝnzat.
Do you think it is better to leave this information out?

karljonaskarlsson: I only include information about faith if there are reasons to be certain, but I honestly don't know if this is the best practice :) how do you do it, @CarstenHoffmannMarburg ?

CarstenHoffmannMarburg: I am rather cautious and usually do not give faith unless there is a clear reason, for example if a person has a position in the church hierarchy. But other encoders in our team use EOTC more often. I think that the possession of a burial book or other religious literature is an indicator of the faith of the owner, and therefore it is legitimate to keep EOTC in the record here, although I think it may also be a bit ambiguous in principle, as people may own books for reasons other than their intended use in service, for example as a valuable or prestigious item, or for personal study. It is up to you to decide whether this is enough to code a person as an EOTC.

MassimoVilla: Better to leave this information out, then. There are no indications that these persons belong to an ecclesiastical hierachy.

karljonaskarlsson (approved):

CarstenHoffmannMarburg (approved):

CarstenHoffmannMarburg:
I think, we have a server down at the moment. Please do not merge, before checking that the bm website is online!

PR #995, opened 2025-01-10
Owners in BLorient4943 and BLorient4944

CarstenHoffmannMarburg:
These are the owners of a magic scroll, that was seperated in several pieces, two of them are kept in the British Library being BLorient4943 and BLorient4944.


thea-m (approved):
Great, thank you!

PR #999, opened 2025-01-20
Magda dd 034 16 01 2025

MagdaKrzyz:


eu-genia (approved):

new/PRS14662TsaggaLaEgziabeher.xml
```diff
@@ -0,0 +1,61 @@
...
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ፀጋ፡ ለእግዚአብሔር፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Ḍaggā Laʾǝgziʾabǝḥer</persName>
+                    <faith type="EOTC"/>
+                    <occupation type="ecclesiastic">scribe</occupation>
```
CarstenHoffmannMarburg, line 51: It is not an ecclesiastic occupation as for instance abbot or deacon. I would not even define it as literary occupation, which should be reserved to researchers, authours or professional scribes. In his case, he is the scribe of a certain book, what gives him a role concerning the production of this item to be encoded in a relation (as you already did), but not an (permanent) occupation (= profession).

new/PRS14663Yehdago.xml
```diff
@@ -0,0 +1,61 @@
...
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ይሕደጎ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Yǝḥdago</persName>
+                    <persName type="alt"><roleName type="title">ʾabeto</roleName>Yǝḥdago</persName>
+                    <faith type="EOTC"/>
+                    <occupation type="other"/>
```
CarstenHoffmannMarburg, line 52: I would recommend to delete occupation, if there is no information available.

CarstenHoffmannMarburg (changes requested):

new/PRS14662TsaggaLaEgziabeher.xml
MagdaKrzyz: I agree

new/PRS14663Yehdago.xml
MagdaKrzyz: ok

new/PRS14662TsaggaLaEgziabeher.xml
thea-m: See also https://github.com/BetaMasaheft/Documentation/issues/1589

MagdaKrzyz: Sorry, I have agreed by mistake. We do not know who are professional scribes and who are not. Most of the scribes are peaseants working on the field. If someone says "I copied a book X" he is a scribe for me by occupation. Otherwise you would have in this filed only the authors of the chronicles. 
In this particular case I can indeed delete the occupation because the hand is very poor and he was perhaps not a scribe but rather a trainee.

CarstenHoffmannMarburg (approved):

PR #1014, opened 2025-02-13
PRS BNFabb68 (a10)

CarstenHoffmannMarburg:
I created PRS records from one addition in BNFabb68 according to our discussion in https://github.com/BetaMasaheft/Documentation/issues/2828?notification_referrer_id=NT_kwDOAmbNy7QxNDc2NzA3Mzg2OTo0MDI5MTc4Nw#event-16298200552.

### DIFF
--- a/new/PRS14701Ayako.xml
+++ b/new/PRS14701Ayako.xml
@@ -47,7 +47,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                 <person sex="1">
                     <persName xml:lang="gez" xml:id="n1">አያኮ፡</persName>
                     <persName xml:lang="gez" type="normalized" corresp="#n1">ʾAyāko</persName>
-                    <occupation type="military">ya-ḥaḍe zallān</occupation>
+                    <occupation type="other">ya-ḥaḍe zallān</occupation>
                     <floruit notBefore="1730" notAfter="1848">18th/19th century, but not before 1730 and not after 1848.</floruit>
                 </person>
                 <listRelation>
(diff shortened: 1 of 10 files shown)


eu-genia (approved):

nafisa-valieva (approved):
Looks good for me! Thank you! I think it would be useful to have titles/occupation in fidal too.

MagdaKrzyz (approved):

new/PRS14701Ayako.xml
```diff
@@ -0,0 +1,59 @@
...
+            <change who="CH" when="2025-02-13">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">አያኮ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">ʾAyāko</persName>
+                    <occupation type="military">ya-ḥaḍe zalān</occupation>
```
MagdaKrzyz, line 50: Not "zalān" but "zallān", the meaning is rather unclear in this context. It means normally "nomad". Did they indeed served as militaries for the king? Unclear

CarstenHoffmannMarburg: I think, that ḥaḍe is a title for the Emperor and therefore ya-ḥaḍe zallān probably is a military occupation. It is necessary to define type in occupation. If we do not want to see it as a military occupation, we can rather define it as 'other' in occupation or as a title or delete this line altogether. Which option do you prefer?

new/PRS14701Ayako.xml
```diff
@@ -47,7 +47,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                 <person sex="1">
                     <persName xml:lang="gez" xml:id="n1">አያኮ፡</persName>
```
MagdaKrzyz, line 48: Should be ʾAyyāko; most probably, as you suggested they were King's soldiers who did not have a fixed place of living but moved together with the court.

PR #1022, opened 2025-02-25
Create PRS14714WaldaAbiya.xml

MagdaKrzyz:


new/PRS14714WaldaAbiya.xml
```diff
@@ -45,19 +45,18 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
...
-                    <persName xml:lang="gez" type="normalized" corresp="#n1"/>
+                    <persName xml:lang="gez" xml:id="n1">ወልደ፡ ዓቢየ፡ እግዚእ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Walda ʿĀbiya ʾƎgziʾ</persName>
                     <faith type="EOTC"/>
-                    <occupation type="other"/>
                     <nationality type="Ethiopia"/>
                     <birth/>
                     <death/>
-                    <floruit/>
+                    <floruit notBefore="1820" notAfter="1980"/>
```
CarstenHoffmannMarburg, line 54: Please express the date also explicitely:
<floruit notBefore="1820" notAfter="1980">18th/20th century, but not before 1820 and not after 1980</floruit>

CarstenHoffmannMarburg (changes requested):

eu-genia (approved):

new/PRS14714WaldaAbiya.xml
MagdaKrzyz: ok

CarstenHoffmannMarburg (approved):

new/PRS14714WaldaAbiya.xml
CarstenHoffmannMarburg: Sorry, I wanted to say "19th/20th century, but not before 1820 and not after 1980".

MagdaKrzyz: @karljonaskarlsson @DenisNosnitsin1970 @nafisa-valieva 
I am not sure if we need this string since there will be written 1820–1980. Is this our usual practice? I am asking because in the previous records I have not written in this way and I was not corrected. 
I would not write "19th/20th c." because a slash means "or", rather "19th–20th c."
If we want to be precise perhaps sth. like "Sometime between the 19th and 20th centuries, but ….."

nafisa-valieva (approved):

new/PRS14714WaldaAbiya.xml
nafisa-valieva: I see no problem. From what Denis told me, there is a certain flexibility, we try to be consistent within one catalogue, but we do not follow any unified way.

MagdaKrzyz: "Is this our usual practice? I am asking because in the previous records I have not written in this way and I was not corrected." - sorry for imprecision, I mean here whether we write the whole string as Carsten suggests.

CarstenHoffmannMarburg: As you decide. I already gave my approval.

CarstenHoffmannMarburg (approved):

DenisNosnitsin1970 (approved):

CarstenHoffmannMarburg (approved):

PR #1053, opened 2025-03-21
PRS (owners) in BLorient11713

CarstenHoffmannMarburg:


new/PRS14751ElleneNeg.xml
```diff
@@ -0,0 +1,58 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="PRS14751ElleneNeg" xml:lang="en" type="pers">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title>ʾƎllenǝ Nǝgǝśt</title>
```
karljonaskarlsson, line 8: Final schwa is very marginal in Geez phonotactics, so if there aren't extremely strong arguments for something else, I would transliterate "ʾƎllen"

karljonaskarlsson (approved):
Looks good!

new/PRS14751ElleneNeg.xml
CarstenHoffmannMarburg: I was surprised by St. Strelcyns translitteration as well. I assume, that he thinks, that the standard form is ʾƎlleni or ʾƎllene with the last vowel weakened and centralised to ʾƎllenǝ. 

I can see only one comparable case with the same name (PRS3737ellen), without a vowel in the end. Therefore, I will go with @karljonaskarlsson and delete the ǝ.

nafisa-valieva (approved):

PR #1055, opened 2025-03-24
PRSs in UUB O Etiop. 9

karljonaskarlsson:


CarstenHoffmannMarburg (approved):

new/PRS14760KenfaGabreel.xml
```diff
@@ -0,0 +1,59 @@
...
+            <change who="JK" when="2025-03-24">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ክንፈ፡ ገብርኤል፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Kənfa Gabrəʾel</persName>
+                    <floruit from="1700" to="1950">1700–1950</floruit>
```
MagdaKrzyz, line 50: Just a question: why such a big span of flourit?

MagdaKrzyz (approved):

new/PRS14760KenfaGabreel.xml
karljonaskarlsson: He appears in a later note in a manuscript for which we don't have images, so I took the earliest possible date as the terminus post quem and the approximate arrival in Sweden as the ante quem. It's quite a large span, I agree, but at least it gives us a rough idea :)

MagdaKrzyz: Alles klar!

PR #1066, opened 2025-04-01
Update

MagdaKrzyz:
I have to remerge the files.


thea-m (approved):

eu-genia (approved):

MagdaKrzyz:
@eu-genia 
There is a problem with visualizing relations: in 6 files only two relations have been visualized. I cannot see the mistake. 
Here is the visualization of PRS14773WaldaMalak:

Image

Here is my encoding:

Image

The same mistake appears with PRS14758TaklaMikael (the most confused visualization), 

PRS14759GabraEgzi, PRS14771GabraMikael, PRS14772NasaraAb, PRS14774GabraMadhen

eu-genia:
This is because lists are currently not generated automatically but manually and I cannot do that on the fly

eu-genia:
No actually the list is updated but you have to open the records (manuscripts) that point to the names for them to be written to the list. I can see the relations all right and I did not do anything at all.

PR #1069, opened 2025-04-02
Nv 2025.03.26.hadgu

nafisa-valieva:


new/PRS14767Hadgu.xml
```diff
@@ -0,0 +1,74 @@
...
+        </profileDesc>
+        <revisionDesc>
+            <change who="NV" when="2025-03-26">Created entity</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">Ḥadgu</persName>
```
thea-m, line 48: Fidal is missing

thea-m (changes requested):
Very nice!

new/PRS14767Hadgu.xml
nafisa-valieva: > Fidal is missing

we do not have it, but I reconstructed it  :-)

thea-m (approved):
It's also ok to only write the transcription if there is any uncertainty, in this case an explicit statement about the situation would be helpful

PR #1074, opened 2025-04-04
Update

MagdaKrzyz:


new/PRS14780NabiyaLeul.xml
```diff
@@ -0,0 +1,64 @@
...
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ነቢየ፡ ልዑል፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Nabiya Lǝʿul</persName>
+                    
+                    <persName><roleName type="title">ʾabbā</roleName> </persName>
+                    <faith type="EOTC"/>
+                    <nationality type="Ethiopia"/>
+                    <occupation>monk</occupation>
```
thea-m, line 54: type="ecclesiastic"?

thea-m (approved):
Great, thank you!

new/PRS14782TawaldaMadhen.xml
```diff
@@ -0,0 +1,61 @@
...
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ተወልደ፡ መድኅን፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Tawalda Madḫǝn</persName>
+                    <faith type="EOTC"/>
+                    <nationality type="Ethiopia"/>
+                    <floruit notBefore="1800" notAfter="1950"/>
+                </person>
+                <listRelation>
+                    <relation name="saws:hasOwned" active="PRS14782TawaldaMadhen" passive="ESdd025"/>
+                    <relation name="snap:KinOf" active="PRS14782TawaldaMadhen" passive="PRS14781GabraMikael"/>
```
karljonaskarlsson, line 56: Just out of curiosity, what is the evidence for this undefined "kinship"? I never used it, because it was always clear that one person was the mother or father or brother or sister etc. of the other.

karljonaskarlsson (approved):
Excellent!

new/PRS14780NabiyaLeul.xml
MagdaKrzyz: done

new/PRS14782TawaldaMadhen.xml
MagdaKrzyz: I have used this relation because I do not know what is the actual relation between the people, it is not said. I assume that they belong to the same family as they are both mentioned in supplications and their names were written by the same scribe. I would bet that only relatives would be mentioned together.

karljonaskarlsson: I understand. My presumption has been that in such cases, they will appear on each other's pages as "attested in the same context", but maybe it's good to mark it like this as well. I can't find a definition of when "kin" is meant to be used :)

MagdaKrzyz: Hmm... your proposal is even better because the kinship is only a conjecture. I will change it. Thanks!

MagdaKrzyz: I was to quick to answer. It seems that I have not understand you well. Is there a relation: "attested in the same context"? Or you mean that the "kin" would imply "attested in the same context"?
Bin confused.

karljonaskarlsson: As far as I remember (but we are not supposed to search at the moment, so I can't show you an example), there is a box on persons' pages in the app which shows "Attestations", and there is automatically shows other people attested in the same context. As far as I remember… we need to check when it's possible.

MagdaKrzyz: good

PR #1090, opened 2025-05-07
PRS in BLorient11339

CarstenHoffmannMarburg:
For Reverend and Bishop Dauglish see https://en.wikipedia.org/wiki/John_Dauglish.


nafisa-valieva (approved):
Thank you!

CarstenHoffmannMarburg:
> Thank you, very nice! If wikipedia is the source for the Dauglish entry, I think it should be included in the record explicitly.

Do you recommend to create a zotero entry and tag for the wikipedia page on Dauglish?

I think, there are very many PRS and LOC records using data from wikipedia without mention.

thea-m:
A quick solution would be to mention the page in the change element or in a `<note>`, though probably a zotero entry would be preferable

nafisa-valieva (approved):

thea-m (approved):

PR #1100, opened 2025-05-16
Add change records to PRS records_num2_CH

CarstenHoffmannMarburg:
cf. https://github.com/BetaMasaheft/Documentation/issues/2976

### DIFF
--- a/new/PRS14690KeflaKrestos.xml
+++ b/new/PRS14690KeflaKrestos.xml
@@ -39,6 +39,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         </profileDesc>
         <revisionDesc>
             <change who="CH" when="2025-02-12">Created entity</change>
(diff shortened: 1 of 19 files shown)


MagdaKrzyz (approved):
I approve but I think if we work as a team we do not have to mention who has given additional suggestions. I you forget to add my name in changes, feel free not to correct it.

nafisa-valieva (approved):

nafisa-valieva:
> I approve but I think if we work as a team we do not have to mention who has given additional suggestions. I you forget to add my name in changes, feel free not to correct it.

Only because Eugenia specified that the reviewing process must be acknowledged did we start to be even more precise in telling who reviewed it. Ultimately, why not credit the reviewers or what it means, otherwise acknowledging the reviewing process. Before reviewing process was not mentioned at all.

CarstenHoffmannMarburg:
I see that this way of encoding the reviewers provides more precise information than has been given before. However, I need to go through all the records to add a second change element. Adding the names of the reviewers to this workflow is not a problem for me.

PR #1112, opened 2025-05-22
Update PRS14872Kiyaru.xml

MagdaKrzyz:
Kiyaru is stubborn. I have to remerge him. Mahamed gave up.


MagdaKrzyz:
Kiyaru gave up too. Thank you for swift reply!

PR #1156, opened 2025-07-07
Dr 25 06 anlcr21

thea-m:


MagdaKrzyz (approved):
All wonderful! I have only one question concerning <flourit>. Is it enough if we encode the dates? I can see that some people prefer to write down the dates between the coding some do not do that. I was corrected many times when I failed to do that.
So in your case: `<floruit notBefore="1800" notAfter="1949">1800-1949</flourit>`

thea-m:
I think it is sometimes preferred for visualisation. I personally don't really mind. (See for example https://betamasaheft.eu/persons/PRS13298WaldaA/main, with visualisation of    `<floruit notBefore="1700" notAfter="1800"/>`). I've added them now.

PR #1167, opened 2025-07-21
PRS Aeth e 35

karljonaskarlsson:
Created new person for Aeth e 35 and corrected minor things in transcription


thea-m (approved):
Great, thank you! The impulse is laudable, but I think there is no need of correct records that were apparently converted from another format and don't seem to adhere to scientific transcription at all

MassimoVilla (approved):

PR #1177, opened 2025-08-10
Update PRS13949Qalamentos.xml

DenisNosnitsin1970:
Redistribution/correction of Qalementos IDs.


thea-m (approved):
Is there another ID for the person mentioned in 4 and 6?

PR #1205, opened 2025-09-17
Two PRS for BLorient12913H (Degen and Sherborn)

CarstenHoffmannMarburg:
Two owners of BLorient12913H.

For Edward Degen, I did not find Switzerland as a type in <nationality> (cf. https://github.com/BetaMasaheft/Documentation/issues/3139).


thea-m (approved):
Switzerland might also be added as a type, it might be needed again

PR #1216, opened 2025-09-26
PRS in ANLcr90

CarstenHoffmannMarburg:
Two names of people upon whom the scribe of the Vorlage of ANLcr90 wished to invoke the blessing of the saint ʾAnānyā and who are mentioned in the copy  (ANLcr90). It is not possible to narrow down the floruit, because the Gadla ʾAnānyā is probably very old and the manuscript is very new (29th century).


MagdaKrzyz:
Not only new but futuristic! (29th century) !

MagdaKrzyz (approved):

PR #1217, opened 2025-09-29
Owner and scribe of BLorient13309

CarstenHoffmannMarburg:
Scribe of p2 and owner of the composite manuscript (for the latter see https://github.com/BetaMasaheft/Documentation/issues/3158).


CarstenHoffmannMarburg:
For the scribe of p1 see https://github.com/BetaMasaheft/Persons/pull/1214.

PR #1347, opened 2026-04-01
new records

thea-m:


new/PRS15344WalattaTH.xml
```diff
@@ -0,0 +1,60 @@
...
+    <text>
+        <body>
+            <listPerson>
+                <person sex="2">
+                    <persName xml:lang="gez" xml:id="n1">ወለተ፡ ተክለ፡ ሃይማኖት፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Walatta Takla Hāymānot</persName>
+                    <floruit when="1700" evidence="1799">eighteenth century</floruit>
+                </person>
+                <listRelation>
+                    <relation name="lawd:hasAttestation" active="PRS15344WalattaTH" passive="BDLaethd23"/>
```
MassimoVilla, line 53: Do we add here a sentence inside `<desc>`, like "She is mentioned in several scribal supplications..." or "Her name is mentioned on fol. ..." and similar? Perhaps it's optional

thea-m: Yes, it would be nice but is not necessary here

PR #1364, opened 2026-05-21
Magda makm 091 persons

MagdaKrzyz:


e-schwarz (approved):

new/PRS15399Walda.xml
```diff
@@ -0,0 +1,63 @@
...
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Walda</persName>
+                    <persName>
+                        <roleName type="title">ʾabbā</roleName>
+                    </persName>
+                    <faith type="EOTC"/>
+                    <occupation type="ecclesiastic">monk</occupation>
+                    <floruit notBefore="1800">Sometime after 1800.</floruit>
+                </person>
+                <listRelation>
+                    <relation name="saws:hasOwned" active="PRS15399Walda" passive="ESmakm091"><desc>His name is incomplete. It appears in a brief, carelessly written ownership note in <ref type="mss" corresp="Esmakm#a5">Additio 5 of Ms. MAKM-091</ref>.</desc></relation>
```
CarstenHoffmannMarburg, line 58: I do not think that names as incomplete as this one should be registered.

MagdaKrzyz: I registered, done. What to do, there is an ownership note.

CarstenHoffmannMarburg: I see, it was created and so it can stay. However I think for the future, it is sufficient to mention the situation in the manuscript record. It is no help to look up a name that is mostly lost.

PR #1391, opened 2026-06-18
Two Italian officers in ANLcr66

CarstenHoffmannMarburg:
According to our discussion in https://github.com/BetaMasaheft/Documentation/issues/3344.

### DIFF
--- a/new/PRS15459Celoria.xml
+++ b/new/PRS15459Celoria.xml
@@ -1,0 +1,62 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="PRS15459Celoria" xml:lang="en" type="pers">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title>Candido Celoria</title>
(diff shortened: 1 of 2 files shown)


new/PRS15459Celoria.xml
```diff
@@ -0,0 +1,62 @@
...
+            <change who="CH" when="2026-06-18">Created record</change>
+            <!--  -->
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1" sameAs="wd:Q140275416">
+                    <persName><forename>Candido</forename> <surname>Celoria</surname></persName>
+                    <persName><roleName type="rank">tenente</roleName></persName>
```
MagdaKrzyz: I would opt for adding an English equivalent in parenthesis

MagdaKrzyz (approved):

MagdaKrzyz:
I see the issue about the two Italians only now. Sorry, I have overlooked it before. You have not added much to their profiles from what is available on them. At lease it would be good to have a relation: how they were connected to each other. There is also sth. where they live or were born.

MagdaKrzyz:
I was too fast to approve and you too fast to merge. I think you could do better with the files

CarstenHoffmannMarburg:
> There is also sth. where they live or were born.

Having had a short glance on the sources, I have not seen any information about where and when Tenente Celoria and Generale have been born and died.

MagdaKrzyz:
But there is info that one was under the other in the army. This is why they are both registered as the owners of the ms. Right?

CarstenHoffmannMarburg:
> But there is info that one was under the other in the army. This is why they are both registered as the owners of the ms. Right?

The lieutenant gave it to his general.

MagdaKrzyz:
And as it is said somewhere the general was the liutentant's superior. I would create a relation between the people "professionalRelation" or sth. that we have

PR #1395, opened 2026-06-25
Persons in Borgianus 13

MassimoVilla:
Dear Magda, these are the scribe and a secondary owner of our Borgianus 13.
Would you approve them (if if you think that everything is ok)? 
The full description of the manuscript is also coming.


MagdaKrzyz (approved):
Great!

PR #1400, opened 2026-07-01
PRS in ANLcr69 (owner and copier)

CarstenHoffmannMarburg:


new/PRS15481WaldaTans.xml
```diff
@@ -0,0 +1,59 @@
...
+            <!--  -->
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ወልደ፡ ተንሣኤ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Walda Tanśāʾe</persName>
+                    <floruit when="1900">He was mentioned as the owner of a manuscript in a colophon dated to 1900 A.D.</floruit>
```
MagdaKrzyz: He IS mentioned

new/PRS15482WaldaGab.xml
```diff
@@ -0,0 +1,59 @@
...
+            <!--  -->
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ወልደ፡ ገብርኤል፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Walda Gabrǝʾel</persName>
+                    <floruit when="1900">He was mentioned as the copier of a manuscript in a colophon dated to 1900 A.D.</floruit>
```
MagdaKrzyz: IS mentioned 
of any ms or of THE ms?

MagdaKrzyz (changes requested):

new/PRS15482WaldaGab.xml
CarstenHoffmannMarburg: He is mentioned in ANLcr69

new/PRS15481WaldaTans.xml
```diff
@@ -48,7 +48,7 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
                 <person sex="1">
                     <persName xml:lang="gez" xml:id="n1">ወልደ፡ ተንሣኤ፡</persName>
                     <persName xml:lang="gez" type="normalized" corresp="#n1">Walda Tanśāʾe</persName>
-                    <floruit when="1900">He was mentioned as the owner of a manuscript in a colophon dated to 1900 A.D.</floruit>
+                    <floruit when="1900">1900 A.D.</floruit>
```
MagdaKrzyz: You have deleted the info on flourit, but it was ok. I just wanted to point out that the past form "was" is wrong. He IS still mentioned there

MagdaKrzyz (approved):

new/PRS15481WaldaTans.xml
CarstenHoffmannMarburg: If you think, that the info is clear and helpful, I can restore it.

MagdaKrzyz: As you wish. Either you restore it to indicate that he flourished not only in the year 1900 but simply that the colophon is from that year, or you say sth. like Around 1900. 

In fact you do not need A.D. because it is A.D. by default.

PR #1402, opened 2026-07-03
Create PRS15488Longinus.xml

e-schwarz:


new/PRS15488Longinus.xml
```diff
@@ -53,14 +53,14 @@ type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
         <body>
             <listPerson>
                 <person sameAs="Q319947" sex="1">
-                    <persName xml:lang="gez" xml:id="n1"/>
+                    <persName xml:lang="gez" xml:id="n1">ለንጊኖስ፡</persName>
```
MagdaKrzyz: I think the main name should be the one in Latin since he is not an Ethiopian personage.

MagdaKrzyz (changes requested):

new/PRS15488Longinus.xml
e-schwarz: done!

new/PRS15488Longinus.xml
```diff
@@ -0,0 +1,71 @@
+<?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+schematypens="http://relaxng.org/ns/structure/1.0"?><?xml-model href="https://raw.githubusercontent.com/BetaMasaheft/schema/master/tei-betamesaheft.rng" 
+type="application/xml" schematypens="http://purl.oclc.org/dsdl/schematron"?>
+<TEI xmlns="http://www.tei-c.org/ns/1.0" xml:id="PRS15488Longinus" xml:lang="en" type="pers">
+    <teiHeader>
+        <fileDesc>
+            <titleStmt>
+                <title>Langinos</title>
```
CarstenHoffmannMarburg: If we prefer the English name, we should also have 'Longinus' in the title.

CarstenHoffmannMarburg (changes requested):

new/PRS15488Longinus.xml
e-schwarz: done

MagdaKrzyz (approved):

CarstenHoffmannMarburg (approved):

PR #1421, opened 2026-07-31
Create PRS15516KeflaYaekob.xml

e-schwarz:


new/PRS15516KeflaYaekob.xml
```diff
@@ -0,0 +1,59 @@
...
+        <revisionDesc>
+            <change who="ESh" when="2026-07-31">Created record</change>
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1">
+                    <persName xml:lang="gez" xml:id="n1">ክፍለ፡ ያዕቆብ፡</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1">Kǝfla Yāʿǝqob</persName>
```
CarstenHoffmannMarburg: I think, we normally transcribe ያዕቆብ with Yāʿqob. 

Here and above in title.

CarstenHoffmannMarburg (changes requested):

new/PRS15516KeflaYaekob.xml
e-schwarz: done

CarstenHoffmannMarburg (approved):

PR #1428, opened 2026-08-13
Create PRS15529Santoli for BFmartini1

CarstenHoffmannMarburg:


new/PRS15529Santoli.xml
```diff
@@ -0,0 +1,61 @@
...
+            <change who="CH" when="2026-08-13">Created record</change>
+            <!--  -->
+        </revisionDesc>
+    </teiHeader>
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1" sameAs="wd:Q3927953">
+                    <persName xml:lang="gez" xml:id="n1">Quinto Santoli</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1"/>
```
e-schwarz: Italian, not Geez

e-schwarz: And we do not need the normalized version, I think, but the encoding of the first and the last names instead

new/PRS15529Santoli.xml
```diff
@@ -0,0 +1,61 @@
...
+        </encodingDesc>
+        <profileDesc>
+            <creation/>
+            <abstract>
+                <p/>
+            </abstract>
+            <langUsage>
+                <language ident="en">English</language>
+                <language ident="gez">Gǝʿǝz</language>
+                <language ident="it">Italian</language>
```
e-schwarz: I do not think we need Geez in this record

e-schwarz (changes requested):

new/PRS15529Santoli.xml
```diff
@@ -0,0 +1,61 @@
...
+        <body>
+            <listPerson>
+                <person sex="1" sameAs="wd:Q3927953">
+                    <persName xml:lang="gez" xml:id="n1">Quinto Santoli</persName>
+                    <persName xml:lang="gez" type="normalized" corresp="#n1"/>
+                    <occupation type="literary">Librarian and historian</occupation>
+                    <nationality type="Italy"/>
+                    <birth when="1875-07-10">10 July 1875</birth>
+                    <death when="1959-09-28">28 September 1959</death>
+                    <floruit notBefore="1900" notAfter="1959">Twentieth century until 1959</floruit>
```
MagdaKrzyz: any relation? Where does he appear?

MagdaKrzyz (changes requested):

new/PRS15529Santoli.xml
CarstenHoffmannMarburg: Correct, but we do not need to mark the language and the xml:id in this case at all.

new/PRS15529Santoli.xml
CarstenHoffmannMarburg: According to Gianfrancesco Lusini and Massimo Villa, some marginal notes in BFmartini1, BFmartini3 and BFmartini4 are probably by his hand.

CarstenHoffmannMarburg: The user can find his attestations by clicking on the "Show attestation" button below the record. I prefer to only include relationships for individuals explicitly named as owners, scribes, or, where applicable, their relatives—not for scholars, catalogers, or librarians, whose attestations are often abundant in manuscripts and it would be too cumbersome to describe them all in detail.

grafik

new/PRS15529Santoli.xml
```diff
@@ -0,0 +1,58 @@
...
+    <text>
+        <body>
+            <listPerson>
+                <person sex="1" sameAs="wd:Q3927953">
+                    <persName><forename>Quinto</forename> <surname>Santoli</surname></persName>
+                    <occupation type="literary">Librarian and historian</occupation>
+                    <nationality type="Italy"/>
+                    <birth when="1875-07-10">10 July 1875</birth>
+                    <death when="1959-09-28">28 September 1959</death>
+                    <floruit notBefore="1900" notAfter="1959">Twentieth century until 1959</floruit>
```
MagdaKrzyz: But here you need the starting point: From the beginning of the 20th century until 1959

new/PRS15529Santoli.xml
MagdaKrzyz: I understand

e-schwarz (approved):

MagdaKrzyz (approved):

